PR #53, opened 2018-10-03
slight change for nxt_memzero recent commit

devnexen:
explicit_bzero is a void function as bzero.
If compiler optimisations defeaters is not available in the system,
providing a basic implementation not relying on architecture dependent
memory barriers which would weighten in the existent codebase.


xeioex:
```

devnexen:
OK let s go for the while loop version ;-)

xeioex:
That was Igor's request:) 
In njs codebase, this is quite idiomatic code.

xeioex:
Resolved in ad92be8ec6ea48dff4cdea1c0e8511f86631a257.

PR #223, opened 2019-09-22
timing safe string comparison proposal.

devnexen:
Mainly useful for hashes comparison.


thresheek:

ms-jpq:
might be better UX to roll this into the `crypto` module instead of a string builtin?

PR #342, opened 2020-10-09
Add TypeScript types for modules

jirutka:
**Questions:**

- What is type `Buffer` – [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) or some custom type?
- Where should be used `NjsStringLike` and where `string`?

Patch: https://patch-diff.githubusercontent.com/raw/nginx/njs/pull/342.patch

### DIFF
--- a/src/ts/fs.d.ts
+++ b/src/ts/fs.d.ts
@@ -1,0 +1,384 @@
+/// <reference path="njs_core.d.ts" />
+
+declare module "fs" {
+
+    /**
+     * File system flag that controls opening of a file.
+     *
+     * - `'a'`   - Open a file for appending. The file is created if it does not exist.
+     * - `'ax'`  - The same as `'a'` but fails if the file already exists.
+     * - `'a+'`  - Open a file for reading and appending. If the file does not exist, it will be created.
+     * - `'ax+'` - The same as `'a+'` but fails if the file already exists.
+     * - `'as'`  - Open a file for appending in synchronous mode. If the file does not exist, it will be created.
+     * - `'as+'` - Open a file for reading and appending in synchronous mode. If the file does not exist, it will be created.
+     * - `'r'`   - Open a file for reading. An exception occurs if the file does not exist.
+     * - `'r+'`  - Open a file for reading and writing. An exception occurs if the file does not exist.
+     * - `'rs+'` - Open a file for reading and writing in synchronous mode. Instructs the operating system to bypass the local file system cache.
+     * - `'w'`   - Open a file for writing. If the file does not exist, it will be created. If the file exists, it will be replaced.
+     * - `'wx'`  - The same as `'w'` but fails if the file already exists.
+     * - `'w+'`  - Open a file for reading and writing. If the file does not exist, it will be created. If the file exists, it will be replaced.
+     * - `'wx+'` - The same as `'w+'` but fails if the file already exists.
+     */
+    export type OpenMode = "a" | "ax" | "a+" | "ax+" | "as" | "as+" | "r" | "r+" | "rs+" | "w" | "wx" | "w+" | "wx+";
+
+    export type FileEncoding = BufferEncoding;
+
+    /**
+     * Valid types for path values in "fs".
+     */
+    export type PathLike = string | Buffer;
+
+    /**
+     * A representation of a directory entry - a file or a subdirectory.
+     *
+     * When `readdirSync()` is called with the `withFileTypes` option, the resulting array contains
+     * `fs.Dirent` objects.
+     */
+    export interface Dirent {
+        /**
+         * @returns `true` if the object describes a block device.
+         */
+        isBlockDevice(): boolean;
+        /**
+         * @returns `true` if the object describes a character device.
+         */
+        isCharacterDevice(): boolean;
+        /**
+         * @returns `true` if the object describes a file system directory.
+         */
+        isDirectory(): boolean;
+        /**
+         * @returns `true` if the object describes a first-in-first-out (FIFO) pipe.
+         */
+        isFIFO(): boolean;
+        /**
+         * @returns `true` if the object describes a regular file.
+         */
+        isFile(): boolean;
+        /**
+         * @returns `true` if the object describes a socket.
+         */
+        isSocket(): boolean;
+        /**
+         * @returns `true` if the object describes a symbolic link.
+         */
+        isSymbolicLink(): boolean;
+
+        /**
+         * The name of the file this object refers to.
+         */
+        name: string;
+    }
+
+    type WriteFileOptions = {
+        mode?: number;
+        flag?: OpenMode;
+    };
+
+    type Constants = {
+        /**
+         * Indicates that the file is visible to the calling process, used by default if no mode
+         * is specified.
+         */
+        F_OK: 0;
+        /**
+         * Indicates that the file can be read by the calling process.
+         */
+        R_OK: 4;
+        /**
+         * Indicates that the file can be written by the calling process.
+         */
+        W_OK: 2;
+        /**
+         * Indicates that the file can be executed by the calling process.
+         */
+        X_OK: 1;
+    };
+
+    interface Promises {
+        /**
+         * Asynchronously tests permissions for a file or directory specified in the `path`.
+         * If the check fails, an error will be returned, otherwise, the method will return undefined.
+         *
+         * @example
+         *   import fs from 'fs'
+         *   fs.promises.access('/file/path', fs.constants.R_OK | fs.constants.W_OK)
+         *     .then(() => console.log('has access'))
+         *     .catch(() => console.log('no access'))
+         *
+         * @since 0.3.9
+         * @param path A path to a file or directory.
+         * @param mode An optional integer that specifies the accessibility checks to be performed.
+         *   Defaults to `fs.constants.F_OK`.
+         */
+        access(path: PathLike, mode?: number): Promise<void>;
+
+        /**
+         * Asynchronously appends specified `data` to a file with provided `filename`.
+         * If the file does not exist, it will be created.
+         *
+         * @since 0.4.4
+         * @param path A path to a file.
+         * @param data The data to write.
+         * @param options An object optionally specifying the file mode and flag.
+         *   If `mode` is not supplied, the default of `0o666` is used.
+         *   If `flag` is not supplied, the default of `'a'` is used.
+         */
+        appendFile(path: PathLike, data: NjsStringLike | Buffer, options?: WriteFileOptions): Promise<void>;
+
+        /**
+         * Asynchronously creates a directory at the specified `path`.
+         *
+         * @since 0.4.2
+         * @param path A path to a file.
+         * @param options The file mode (or an object specifying the file mode). Defaults to `0o777`.
+         */
+        mkdir(path: PathLike, options?: { mode?: number } | number): Promise<void>;
+
+        /**
+         * Asynchronously reads the contents of a directory at the specified `path`.
+         *
+         * @since 0.4.2
+         * @param path A path to a file.
+         * @param options A string that specifies encoding or an object optionally specifying
+         *   the following keys:
+         *   - `encoding` - `'utf8'` (default) or `'buffer'` (since 0.4.4)
+         *   - `withFileTypes` - if set to `true`, the files array will contain `fs.Dirent` objects; defaults to `false`.
+         */
+        readdir(path: PathLike, options?: { encoding?: "utf8"; withFileTypes?: false; } | "utf8"): Promise<string[]>;
+        readdir(path: PathLike, options: { encoding: "buffer"; withFileTypes?: false; } | "buffer"): Promise<Buffer[]>;
+        readdir(path: PathLike, options: { encoding?: "utf8" | "buffer"; withFileTypes: true; }): Promise<Dirent[]>;
+
+        /**
+         * Asynchronously returns the contents of the file with provided `filename`.
+         * If an encoding is specified, a `string` is returned, otherwise, a `Buffer`.
+         *
+         * @param path A path to a file.
+         * @param options A string that specifies encoding or an object with the following optional keys:
+         *   - `encoding` - `'utf8'`, `'hex'`, `'base64'`, or `'base64url'` (the last three since 0.4.4).
+         *   - `flag` - file system flag, defaults to `r`.
+         */
+        readFile(path: PathLike): Promise<Buffer>;
+        readFile(path: PathLike, options?: { flag?: OpenMode; }): Promise<Buffer>;
+        readFile(path: PathLike, options: { encoding?: FileEncoding; flag?: OpenMode; } | FileEncoding): Promise<string>;
+
+        /**
+         * Asynchronously computes the canonical pathname by resolving `.`, `..` and symbolic links using
+         * `realpath(3)`.
+         *
+         * @since 0.3.9
+         * @param path A path to a file.
+         * @param options The encoding (or an object specifying the encoding), used as the encoding of the result.
+         */
+        realpath(path: PathLike, options?: { encoding?: "utf8" } | "utf8"): Promise<string>;
+        realpath(path: PathLike, options: { encoding: "buffer" } | "buffer"): Promise<Buffer>;
+
+        /**
+         * Asynchronously changes the name or location of a file from `oldPath` to `newPath`.
+         *
+         * @since 0.3.4
+         * @param oldPath A path to a file.
+         * @param newPath A path to a file.
+         */
+        rename(oldPath: PathLike, newPath: PathLike): Promise<void>;
+
+        /**
+         * Asynchronously removes a directory at the specified `path`.
+         *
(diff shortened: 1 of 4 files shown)


jirutka:
@xeioex, could you please give me some feedback and answer the questions above?

xeioex:
@jirutka 

Sorry for delay, I plan to return to your merge request later.

> What is type Buffer – ArrayBuffer or some custom type?

No, it replicates API for [node.js Buffer](https://nodejs.org/api/buffer.html)

> NjsStringLike

Basically it is a [pseudotype](https://github.com/nginx/njs/blob/master/src/ts/njs_core.d.ts#L41) for TS, which says that some function accepts both string and NjsByteString.

In the future, we plan to get rid of NjsByteString (as a non-standard type) in favour of Buffer.

xeioex:
@jirutka 

Regarding the modules API, we are trying to follow the original API for them, can't we just copy the existing API description for them?

jirutka:
Thanks for the quick response!

> No, it replicates API for node.js Buffer

Okay, I will update the type definitions.

> Basically it is a pseudotype for TS, which says that some function accepts both string and NjsByteString.

I understand that. I just don’t know how to recognize what functions accept/return `NjsByteString` and what functions just plain `string`.

> Regarding the modules API, we are trying to follow the original API for them, can't we just copy the existing API description for them?

You mean Node.js API? You don’t follow them fully, there are some minor differences (e.g. missing functions, some extra arguments not supported, limited set of encodings etc.). It’s more a subset of them. The TS types must to be accurate to be useful.

BTW, I have copied JSDoc comments from http://nginx.org/en/docs/njs/reference.html, not Node.js docs, both due to differences and to avoid potential legal issues (not sure if there are any, just to be sure).

jirutka:
> No, it replicates API for node.js Buffer

Node’s [https://nodejs.org/api/buffer.html](Buffer) has quite a large API surface – does njs support **all** of its (non-deprecated) methods with **all** arguments? What _encodings_ does it support?

xeioex:
@jirutka 
>I just don’t know how to recognize what functions accept/return NjsByteString and what functions just plain string.

yes, that is the point of deprecation of NjsByteString type.

Currently, we basically have two types `string` and `NjsByteString` implemented in a single String.prototype.

The only difference is the way they treat characters. 
NjsByteString - a character is always a single byte.
string - a character is a UTF8 codepoint (this is not the same as UTF16 from the spec, but the difference scope is pretty negligible, only rarely used surrogate pairs are affected). 

String.prototype.toUTF8(), and String.prototype.fromUTF8()  are used to convert in both directions.

xeioex:
@jirutka 

> Node’s Buffer has quite a large API surface – does njs support all of its (non-deprecated) methods with all arguments? What encodings does it support?

We plan to release the documentation for it by the end of this week.
Most is the (non-deprecated) methods are supported.

With the exception of: buf.keys(), buf.entries(), buf.values(), buf.readBigInt64BE() (and other BigInt methods).

The available encodings are: utf8, hex, base64, base64url.

jirutka:
BTW, I haven’t tested it yet, but it should be possible to generate TS types and HTML documentation from TSDoc comments included directly in the `.c` files. This would help keep implementation, types and documentation in-sync. Would you be interested in this approach?

xeioex:
@jirutka 

It would be great if you share a working POC (using a `.c` file and TSDoc).

xeioex:
http://nginx.org/en/docs/njs/reference.html#buffer - Buffer docs.

jirutka:
Okay, I’ve just added the Buffer type to this PR.

jirutka:
If I understand `njs_buffer.c` correctly, `Buffer.prototype.write` returns `number`, but all `write*` methods for numbers return `void`. However, in Node’s `Buffer`, all `write*` methods return `number`. Is this intentional…?

lexborisov:
Hi @jirutka 

This is my mistake, I will fix it on Monday.
Thanks for the report!

lexborisov:
@jirutka 

Try this patch:
```diff
# HG changeset patch
# User Alexander Borisov <alexander.borisov@nginx.com>
# Date 1602919869 -10800
#      Sat Oct 17 10:31:09 2020 +0300
# Node ID b4429c95c89a1b54da3450aef85db96d0f650d23
# Parent  a82f123409b7c0c8b97417e11dffde523018ed30
Fixed returned value for Buffer.prototype.write* functions.

All Buffer.prototype.write* functions must return offset plus
the number of bytes written.

The issue was introduced in 27bb9caf186c.

diff -r a82f123409b7 -r b4429c95c89a src/njs_buffer.c
--- a/src/njs_buffer.c	Tue Oct 13 15:44:33 2020 +0300
+++ b/src/njs_buffer.c	Sat Oct 17 10:31:09 2020 +0300
@@ -1331,7 +1331,7 @@ njs_buffer_prototype_write_int(njs_vm_t 
         break;
     }
 
-    njs_set_undefined(&vm->retval);
+    njs_set_number(&vm->retval, index + size);
 
     return NJS_OK;
 }
diff -r a82f123409b7 -r b4429c95c89a src/test/njs_unit_test.c
--- a/src/test/njs_unit_test.c	Tue Oct 13 15:44:33 2020 +0300
+++ b/src/test/njs_unit_test.c	Sat Oct 17 10:31:09 2020 +0300
@@ -19191,7 +19191,11 @@ static njs_unit_test_t  njs_test[] =
               "        if (size > 1) { rmethod += endianness; wmethod += endianness; };"
               "        var v = 0x7abbccddeeff & (size * 8 - 1);"
               ""
-              "        buf[wgmethod](v, offset, size);"
+              "        var ret = buf[wgmethod](v, offset, size);"
+              "        if(ret !== offset + size) {"
+              "            throw Error(`${wgmethod} returned ${ret}, need ${offset + size}`);"
+              "        }"
+              ""
               "        var gv = buf[rgmethod](offset, size);"
               ""
               "        buf.fill(0);"
```

src/ts/fs.d.ts
```diff
@@ -0,0 +1,379 @@
...
+        /**
+         * Asynchronously computes the canonical pathname by resolving `.`, `..` and symbolic links using
+         * `realpath(3)`.
+         *
+         * @since 0.3.9
+         * @param path A path to a file.
+         * @param options The encoding (or an object specifying the encoding), used as the encoding of the result.
+         */
+        realpath(path: NjsStringLike, options?: { encoding?: "utf8" } | "utf8"): Promise<NjsStringLike>;
+        realpath(path: NjsStringLike, options: { encoding: "buffer" } | "buffer"): Promise<Buffer>;  // TODO: is this correct?
```
jirutka, line 169: Is this correct?

src/ts/fs.d.ts
```diff
@@ -0,0 +1,379 @@
...
+        /**
+         * Synchronously computes the canonical pathname by resolving `.`, `..` and symbolic links using
+         * `realpath(3)`.
+         *
+         * @since 0.3.9
+         * @param path A path to a file.
+         * @param options The encoding (or an object specifying the encoding), used as the encoding of the result.
+         */
+        realpathSync(path: NjsStringLike, options?: { encoding?: "utf8" } | "utf8"): NjsStringLike;
+        realpathSync(path: NjsStringLike, options: { encoding: "buffer" } | "buffer"): Buffer;  // TODO: is this correct?
```
jirutka, line 315: Is this correct?

jirutka:
I’ve polished it a bit and I think that it’s ready.

src/ts/njs_core.d.ts
```diff
@@ -35,11 +37,544 @@ type NjsByteString = string & {
...
+    toJSON(): { type: "Buffer"; data: number[] };
+
+    /**
+     * Decodes this buffer to a string according to the specified character `encoding`.
+     *
+     * @param encoding The character encoding. Defaults to `'utf8'`.
+     * @param start The byte offset to start decoding at. Defaults to `0`.
+     * @param end The byte offset to stop decoding at (not inclusive). Defaults to `buf.length`.
+     */
+    toString(encoding?: BufferEncoding, start?: number, end?: number): NjsStringLike;
```
xeioex, line 429: the retval is always a string here.

NjsStringLike -> string.

src/ts/njs_core.d.ts
```diff
@@ -35,11 +37,544 @@ type NjsByteString = string & {
...
+     * Copies the passed buffer `data` onto a new `Buffer` instance.
+     *
+     * @param data The buffer to copy.
+     */
+    static from(data: Buffer): Buffer;
+    /**
+     * For objects whose `valueOf()` function returns a value not strictly equal to object, returns
+     * `Buffer.from(object.valueOf(), offsetOrEncoding, length)`.
+     *
+     * @param obj An object supporting `Symbol.toPrimitive` or `valueOf()`.
```
xeioex, line 127: `Symbol.toPrimitive` is not supported.

src/ts/njs_core.d.ts
```diff
@@ -35,11 +37,544 @@ type NjsByteString = string & {
...
+     */
+    // FIXME: This documentation is incomplete.
```
xeioex, line 129: `Buffer.from(arrayBuffer[, byteOffset[, length]])` is implemented, but missed in the reference docs (will be fixed).

xeioex (changes requested):

src/ts/njs_core.d.ts
jirutka: Actually, it seems that not even `valueOf` is implemented.

```
>> Buffer.from(new Date())              
Thrown:
TypeError: first argument date is not a string or Buffer-like object
    at Buffer.from (native)
    at main (shell:1)
```

or 

```
>> Buffer.from({ valueOf: () => 'foo' })
Thrown:
TypeError: first argument object is not a string or Buffer-like object
    at Buffer.from (native)
    at main (shell:1)
```

However, there is the code path for an object…

https://github.com/nginx/njs/blob/976ed440e983f56a96630caae9103f1cf80345cf/src/njs_buffer.c#L252-L258

But if I understand `njs_buffer_from_object` correctly, it accepts only one argument instead of 3 (`Buffer.from(object[, offsetOrEncoding[, length]])`):

https://github.com/nginx/njs/blob/976ed440e983f56a96630caae9103f1cf80345cf/src/njs_buffer.c#L267-L370

xeioex: yes, this was missed. @lexborisov  is on it.

drsm:
@xeioex 
@jirutka 

Hi!

Looks like all the paths in `fs` module should be of type ` export type PathLike = string | Buffer` instead of `NjsStringLike`

src/ts/fs.d.ts
xeioex: realpath(path: NjsStringLike, options?: { encoding?: "utf8" } | "utf8"): Promise<string>;
realpath(path: NjsStringLike, options: { encoding: "buffer" } | "buffer"): Promise<Buffer>; // yes, this is correct

src/ts/fs.d.ts
xeioex: realpathSync(path: NjsStringLike, options?: { encoding?: "utf8" } | "utf8"): string;
realpathSync(path: NjsStringLike, options: { encoding: "buffer" } | "buffer"): Buffer; // this is correct

jirutka:
I’ve fixed all the mentioned issues in the `crypto` and `querystring` modules.

It seems that `NjsStringLike` is used basically as in input type only (function parameters), not as the return type, right? If so, the following in the `fs` module should be changed too, right?

```diff
- readFileSync(path: NjsStringLike, options: { encoding?: FileEncoding; flag?: OpenMode; } | FileEncoding): NjsStringLike;
+ readFileSync(path: NjsStringLike, options: { encoding?: FileEncoding; flag?: OpenMode; } | FileEncoding): string;
```

```diff
- readdirSync(path: NjsStringLike, options?: { encoding?: "utf8"; withFileTypes?: false; } | "utf8"): NjsStringLike[];
+ readdirSync(path: NjsStringLike, options?: { encoding?: "utf8"; withFileTypes?: false; } | "utf8"): string[];
```

```diff
- realpathSync(path: NjsStringLike, options?: { encoding?: "utf8" } | "utf8"): NjsStringLike;
+ realpathSync(path: NjsStringLike, options?: { encoding?: "utf8" } | "utf8"): string;
```

```diff
- readFile(path: NjsStringLike, options: { encoding?: FileEncoding; flag?: OpenMode; } | FileEncoding): Promise<NjsStringLike>;
+ readFile(path: NjsStringLike, options: { encoding?: FileEncoding; flag?: OpenMode; } | FileEncoding): Promise<string>;
```

```diff
- readdir(path: NjsStringLike, options?: { encoding?: "utf8"; withFileTypes?: false; } | "utf8"): Promise<NjsStringLike[]>;
+ readdir(path: NjsStringLike, options?: { encoding?: "utf8"; withFileTypes?: false; } | "utf8"): Promise<string[]>;
```

```diff
- realpath(path: NjsStringLike, options?: { encoding?: "utf8" } | "utf8"): Promise<NjsStringLike>;
+ realpath(path: NjsStringLike, options?: { encoding?: "utf8" } | "utf8"): Promise<string>;
```

I’ve also changed `Buffer` module to extend `Uint8Array` (as in Node.js) and accept `Uint8Array` instead of `Buffer` (its subtype) in relevant functions, after I’ve noticed in njs shell that it really behaves in the same way as in Node regarding `Uint8Array`.

xeioex:
> It seems that NjsStringLike is used basically as in input type only (function parameters), not as the return type, right?

@jirutka yes, since 0.4.4 "fs" method ensures that ordinary string are returned (as a part of the plan to eliminate NjsByteString).

jirutka:
Okay, updated: all the path parameters accept `PathLike` (`PathLike = string | Buffer`), functions don’t return `NjsStringLike`, but `string`, `NjsStringLike` is used only for the input data parameters.

I’ve also tagged `String.bytesFrom()` as `@deprecated`.

src/ts/fs.d.ts
```diff
@@ -0,0 +1,384 @@
...
+    interface Promises {
+        /**
+         * Asynchronously tests permissions for a file or directory specified in the `path`.
+         * If the check fails, an error will be returned, otherwise, the method will return undefined.
+         *
+         * @example
+         *   import fs from 'fs'
+         *   fs.promises.access('/file/path', fs.constants.R_OK | fs.constants.W_OK)
+         *     .then(() => console.log('has access'))
+         *     .catch() => console.log('no access'))
```
drsm, line 107: `catch()` => `catch(()`

nginx/ts/ngx_stream_js_module.d.ts
```diff
@@ -66,7 +66,7 @@ interface NginxStreamVariables {
     readonly 'time_iso8601'?: NjsByteString;
     readonly 'time_local'?: NjsByteString;
 
-    [prop: string]: NjsByteString;
+    [prop: string]: NjsByteString | undefined;
```
xeioex, line 69: @jirutka Can you share the exact code snippet that causes the issue?

xeioex:
@jirutka BTW, is there a proper way to reference internal types like Hash? https://github.com/nginx/njs/blob/master/test/ts/test.ts#L87

Something like:
```ts
var h:Hash;

h = crypto.createHash('sha1');
```

nginx/ts/ngx_stream_js_module.d.ts
jirutka: Sure, it’s part of #344:

```sh
git clone -b njs-types-pkg git@github.com:jirutka/njs.git
cd njs/ts
npm install
npm test
```

The problem is that you’re running `tsc` with the defaults – backward compatible settings that are very lax and shouldn’t be used for any new projects.

jirutka:
> @jirutka BTW, is there a proper way to reference internal types like Hash? https://github.com/nginx/njs/blob/master/test/ts/test.ts#L87
>
> Something like:
>
> ```
> var h:Hash;
>
> h = crypto.createHash('sha1');
> ```

Well, `Hash` is exported, so it’s not an internal type, you can just import it:

```ts
import crypto, { Hash } from 'crypto'

const h: Hash = crypto.createHash('sha1')
```

TypeScript will automatically remove named imports of types from the generated ES2015 code.

nginx/ts/ngx_stream_js_module.d.ts
jirutka: Sorry for mixing this commit into this PR, I should have created a new PR for it.

jirutka: I’m gonna move this commit into #344 and close this PR, ’cause it has been already merged, except this one.

jirutka:
Merged in 993e7a8d0a801211bd4ffd4ebb74b1d5ceeb443d, 20c02d0689791517a74f6ddecc94edf6a21c909e, ef46a05f2e2888a1b07433640579db25655d1376, e246283dc90fe251958d576d1ffa67eeb0587d09, fb7ba092ad261ad937eb4f1c38d1f7fd868e7ac2.

PR #344, opened 2020-10-20
Prepare TS types for publishing on npm

jirutka:
Patch: https://patch-diff.githubusercontent.com/raw/nginx/njs/pull/344.patch


xeioex:
`./configure && make ts_test`

```
test/ts/test.ts:1:16 - error TS2307: Cannot find module 'fs'.

1 import fs from 'fs';
...
Found 19 errors.
```

xeioex:
@jirutka 
`./configure && make ts_test`

```
test/ts/test.ts:1:16 - error TS2307: Cannot find module 'fs'.

1 import fs from 'fs';
...
Found 19 errors.
```

jirutka:
This PR is not ready yet.

jirutka:
You can try it now. However, it’s still not ready for merge.

I wanted to ask, would it be acceptable to use `npm install` in `ts_test` to install the specified version of [typescript](https://www.npmjs.com/package/typescript) from npm (and possibly other test dependencies in the future) and link `njs-types` from local directory?

xeioex:
@jirutka 

> I wanted to ask, would it be acceptable to use npm install in ts_test to install the specified version of typescript from npm (and possibly other test dependencies in the future) and link njs-types from local directory?

yes, this is acceptable because it is for developers only, and should be called explicitly (this should not be a dependency of any common targets like njs, test etc).

> njs-types from local directory?

from `build` directory, right? this should be the only modifiable directory.

jirutka:
> yes, this is acceptable because it is for developers only, and should be called explicitly (this should not be a dependency of any common targets like njs, test etc).

Okay, great!

> from `build` directory, right? this should be the only modifiable directory.

Okay, I’ve modified the Makefile in this regards.


It’s almost ready. Please take a look if this approach makes sense to you.

ts/ngx_http_js_module.d.ts
```diff
@@ -39,7 +39,7 @@ interface NginxHeadersIn {
     readonly 'Warning'?: NjsByteString;
     readonly 'X-Forwarded-For'?: NjsByteString;
 
-    readonly [prop: string]: NjsByteString;
+    readonly [prop: string]: NjsByteString | undefined;
```
jirutka, line 42: Discussed in https://github.com/nginx/njs/pull/342#discussion_r508623283

jirutka:
I’ve originally suggested using the version number in format `<njs-version>-<types-rev>` (e.g. `0.4.2-1`) to maintain one-to-one relation between njs and njs-types versions and at the same time allow releasing new revisions of njs-types between njs releases. However, I eventually came to the conclusion that it was a bad idea.

npmjs enforces [SemVer](https://semver.org/) which doesn’t allow four-part version number nor provide post-release suffixes. The version suffix `-<types-rev>` actually denotes a **pre**-release version. I thought that it can be used in this case assuming that you will append this suffix to each released version, i.e. you will not release e.g. `0.4.2` (which has **higer** precedence than `0.4.2-1`, `0.4.2-2`…). The problem is that `-1`, `-2`, … has lower precedence even than `-dev.2`, `.beta.3` etc. And it would be also a bit confusing for the users because `^0.4.2` doesn’t include `0.4.2-1`.

For these reasons, I think that it will be better use a similar approach as [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library). I’ve described it in `ts/README.md`:

> njs-types is typically being released together with njs. Their major and minor release numbers (the first two numbers) are always aligned, but the patch version (the third number) may differ. That's because njs-types may be updated between njs releases and in such case the patch version is incremented.

When `./ts` is copied into `./build/ts`, the version in `package.json` is substituted with version read from `src/njs.h`. This can be overriden using make variable `NJS_TYPES_VER` (e.g. `./configure && make ts_publish NJS_TYPES_VER=0.4.2`).

xeioex:
@jirutka 

Here are my improvements over your work
https://gist.github.com/xeioex/ea94f68579a857e58143e24511b8f6db

I am still working on it (planning to merge and simplify some patches further)

But, I am have an issue right now with modules

```
$ make ts_test
...
test.ts:1:16 - error TS2307: Cannot find module 'fs' or its corresponding type declarations.

1 import fs from 'fs';
                 ~~~~

test.ts:2:16 - error TS2307: Cannot find module 'querystring' or its corresponding type declarations.

2 import qs from 'querystring';
                 ~~~~~~~~~~~~~

test.ts:3:20 - error TS2307: Cannot find module 'crypto' or its corresponding type declarations.

3 import crypto from 'crypto';
                     ~~~~~~~~


Found 3 errors.
```

jirutka:
Well, that’s because you’ve excluded the following changes (`-` is this PR, `+` is yours):

```diff
--- a/ts/ngx_http_js_module.d.ts
+++ b/ts/ngx_http_js_module.d.ts
@@ -1,4 +1,4 @@
-/// <reference path="index.d.ts" />
+/// <reference path="njs_core.d.ts" />
--- a/ts/ngx_stream_js_module.d.ts
+++ b/ts/ngx_stream_js_module.d.ts
@@ -1,4 +1,4 @@
-/// <reference path="index.d.ts" />
+/// <reference path="njs_core.d.ts" />
--- a/ts/njs_shell.d.ts
+++ b/ts/njs_shell.d.ts
@@ -1,4 +1,4 @@
-/// <reference path="index.d.ts" />
+/// <reference path="njs_core.d.ts" />
```

I’ve changed references to `index.d.ts` so the devs using njs-types don’t have to reference two files: one of the entrypoints plus `njs_core.d.ts`.

BTW, why did you remove/didn’t include README.md? This is for npmjs.com, it automatically reads README.md from the package shows it as the package’s description.

I forgot that you have to take the patch and apply it on the mercurial repository, so you cannot easily sync the changes. I should probably stop rebasing commits and just add fixups instead… I’m gonna do it for the last time now just to sync your changes back to this PR. :)

Was the removal of `null` in `NginxHeadersOut` intentional? I’m not sure if it should accept it or not.

xeioex:
@jirutka 

> BTW, why did you remove/didn’t include README.md? This is for npmjs.com, it automatically reads README.md from the package shows it as the package’s description.

it is in progress, I did not get to README.md patch yet (plan to add).

>Was the removal of null in NginxHeadersOut intentional? I’m not sure if it should accept it or not.

yes, I plan to simply drop the test which caused the issue.

jirutka:
> yes, I plan to simply drop the test which caused the issue.

Okay, I’ve synced this change back to this PR. And that was the last rebase.

> The meaning of reordering is to reduce the patches size (by squashing some changes).

Yeah, I understand that.

xeioex:
@jirutka 

Here is my current patchset
https://gist.github.com/xeioex/f62ea58fabb6022359c20e7f28638660

Remaining problems are:

1) `ts/*.json ts/*.d.ts ts/**/*.d.ts` it does not look like a portable make deps syntax (I looks like a globstar bash extention https://unix.stackexchange.com/questions/117826/bash-globstar-matching/315116).

2) moreover `make ts_publish NJS_TYPES_VER=0.4.2`

may not work as expected, if `make ts_test` was already invoked with different NJS_TYPES_VER value.
Because `$NJS_BUILD_DIR/ts/package.json` does not explicitly depend on NJS_TYPES_VER.

NJS_TYPES_VER dependency can be introduced, but maybe the dump targets (no dependencies) are a better option for the task.

jirutka:
The [patchset](https://gist.github.com/xeioex/f62ea58fabb6022359c20e7f28638660) is It’s broken because there’s missing `ts/index.d.ts`.

> ```diff
> # auto/make:
> -$NJS_BUILD_DIR/test/ts/package.json: test/ts/*.json test/ts/*.ts
> +$NJS_BUILD_DIR/test/ts/package.json: ts/*.json ts/*.d.ts ts/**/*.d.ts
>         mkdir -p $NJS_BUILD_DIR/test
>         cp -fr test/ts $NJS_BUILD_DIR/test/
> ```

There should be `test/ts/*.json test/ts/*.ts`, not `ts/*.json ts/*.d.ts ts/**/*.d.ts` – this target copies `./test/ts` into `./build/test/ts`, not `./ts`.

> ```diff
> # test/ts/tsconfig.json:
>          "strictPropertyInitialization": true,
> -        "noImplicitThis": true,
> +          "noImplicitThis": true,
>          "alwaysStrict": true,
> ```

There are two extra spaces.

Just a suggestion (I find the current indentation confusing):

 ```diff
# auto/make:
 $NJS_BUILD_DIR/test/ts/node_modules: \\
-       $NJS_BUILD_DIR/njs-types-\$(NJS_TYPES_VER).tgz \\
-       $NJS_BUILD_DIR/test/ts/package.json
+               $NJS_BUILD_DIR/njs-types-\$(NJS_TYPES_VER).tgz \\
+               $NJS_BUILD_DIR/test/ts/package.json
        cd $NJS_BUILD_DIR/test/ts && \$(NPM) install \\
-       --save-dev file:../../njs-types-\$(NJS_TYPES_VER).tgz
+               --save-dev file:../../njs-types-\$(NJS_TYPES_VER).tgz
        cd $NJS_BUILD_DIR/test/ts && \$(NPM) install
```

* * *

> 1. ts/*.json ts/*.d.ts ts/**/*.d.ts it does not look like a portable make deps syntax (It looks like a globstar bash extention

I don’t have that much experience with writing (portable) Makefiles to disquish between POSIX make and GNU make.
What about his approach? https://github.com/nginx/njs/commit/3f407ce2e4d343ad11d1437b2a1d1415fd12f4be.patch

> moreover make ts_publish NJS_TYPES_VER=0.4.2
> may not work as expected, if make ts_test was already invoked with different NJS_TYPES_VER value.
> Because $NJS_BUILD_DIR/ts/package.json does not explicitly depend on NJS_TYPES_VER.

You’re right. Here is a fix: https://github.com/nginx/njs/commit/4412267a015d5ecdf309cae004f98db4afd19742.patch

**Comment by xeioex** [comment]
@jirutka 

You may also help us in improving the development guide for the end users (possibly for newcomers to the node.js world). 
See http://nginx.org/en/docs/njs/typescript.html.

I am not a JS nor node.js developer. So best practices and good examples for writing njs code using njs-types are welcome.

**Comment by jirutka** [comment]
> You may also help us in improving the development guide for the end users (possibly for newcomers to the node.js world). See http://nginx.org/en/docs/njs/typescript.html.

I’ll look into it after I actually finish the setup of my project using njs.

> For example, what is the proper command to produce JS artefact using npm? (analogous to tsc test.ts)

You just declare it as the `build` [script](https://docs.npmjs.com/cli/v6/using-npm/scripts) in package.json, e.g.:

```jsonc
{ ...
  "scripts": {
    "build": "tsc test.ts"
  },
  ...
}
```

and then run `npm build`.

BTW, `tsc` in `test/ts/` now does not produce `test.js` (which we don’t need here) due to `"noEmit": true` in `tsconfig.json`.

* * *

> Here is my latest patchset: https://gist.github.com/e94a527711e04011c616ccc36025e540

There is still `ts/index.d.ts` missing so it doesn’t work. It should be included in commit `Types: refactored working with TypeScript API description.`.

> Introduced .hgignore.

This commit doesn’t introduce just `.hgignore` file, but also `.gitignore`, so it should be named `Introduced .hgignore and .gitignore.` Speaking of which, it would be good to ignore `/build` and `/Makefile` as well:

```diff
# /.hgignore:
+ ^build/
+ ^Makefile$
```

```diff
# /.gitignore
+ /build/
+ /Makefile
```

> Types: added missing var "console" into njs_shell.

To be perfectly correct, this commit should come _after_ `Types: added tsconfig.json and package.json for type descriptions.` because the default typescript config (used when `tsconfig.json` is not present) includes the browser environment where `console` is already defined which results in error.

> ```diff
> + NJS_TEST_TS_SRCS=$(find test/ts/ -name "*.d.ts" -o -name "*.json")
> ```

This is incorrect, `test/ts/` currently doesn’t contain any `*.d.ts` files, but `*.ts` files (just one actually, `test.ts`).

jirutka:
I’ve updated `compilerOptions.lib` in `tsconfig.json` to match njs [compatibility](http://nginx.org/en/docs/njs/compatibility.html): 
https://github.com/nginx/njs/pull/344/commits/708c204d57ba79691e1438eeb0cd052600cb5f35.patch

xeioex:
@jirutka 

> If you just wanted to avoid pipe in this subshell, you can also write it as $(find test/ts/ -name "*.ts" -o -name "*.json" -exec echo {} +) to produce a space-separated (instead of newline-separated) filenames which can be directly used as dependencies in Makefile.

Actually, this is not needed, because newlines are replaced with spaces in the second [command substitution](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_03) (for njs_test_ts_deps).

xeioex:
@jirutka 
Thank for your contribution. Committed.

jirutka:
You’re welcome!

Why did you release njs-types 0.5.0? The njs-types version should be aligned with the njs version and there’s no 0.5.0 yet. The latest released njs version is 0.4.4 and the types currently reflects this version, so you should release njs-types 0.4.4.
For the pre-release / development versions, use a standard pre-release suffix in format `-<sufix>.<num>`, e.g. `-dev.1`, `-pre.1`, `-alpha.1` etc.
Numeric-only pre-release suffix (`-2`) is valid per SemVer, but it’s generally not recommended – it has unclear semantic. I’ve suggested it just as a workaround explained [here](https://github.com/nginx/njs/pull/344#issuecomment-716049575), but eventually dropped it as a bad idea.

xeioex:
@jirutka 

> Why did you release njs-types 0.5.0? The njs-types version should be aligned with the njs version and there’s no 0.5.0 yet.

This is my bad. I accidentally did `make ts_publish`. I think defining `NJS_TYPES_VER`  by default, was not the best idea. `ts_publish` should check whether NJS_TYPES_VER is defined, and warn the user if it is not.

jirutka:
> I think defining NJS_TYPES_VER by default, was not the best idea.

Yeah, it looks like that.

jirutka:
Merged in [07f3d66..fc08c95](https://github.com/nginx/njs/compare/07f3d66~1..fc08c95).

xeioex:
@jirutka 

> Yeah, it looks like that.

Take a look https://gist.github.com/xeioex/95fe47ae0fce376e18de06674e6dbe50.

jirutka:
> Take a look https://gist.github.com/xeioex/95fe47ae0fce376e18de06674e6dbe50.

This is not good, it would require providing `NJS_TYPES_VER` even for running `ts_test`.

PR #353, opened 2020-10-31
Types: fixed type of subrequest() with detached: true.

jirutka:
Order of method overloading declaration matters.


xeioex:
Merged, thanks.

PR #357, opened 2020-11-24
Types: added definitions for setTimeout, setImmediate and clearTimeout.

jirutka:
Patch: https://github.com/nginx/njs/pull/357.patch

### DIFF
--- a/ts/njs_core.d.ts
+++ b/ts/njs_core.d.ts
@@ -607,3 +607,45 @@ interface NjsProcess {
 }
 
 declare const process: NjsProcess;
+
+/**
+ * A value returned by `setTimeout()` and `setImmediate()` functions. It's an positive integer now,
+ * but this may be changed in future, so it should be treated as an opaque value.
+ */
+type TimerHandle = number & { readonly '': unique symbol };
+
+/**
+ * Schedules the "immediate" execution of the given function after I/O events' callbacks.
+ *
+ * @param callback The function to call.
+ * @param args Optional arguments to pass to the `callback` function.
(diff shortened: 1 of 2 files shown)


ts/njs_core.d.ts
```diff
@@ -606,3 +606,45 @@ interface NjsProcess {
...
+
+/**
+ * A value returned by `setTimeout()` and `setImmediate()` functions. It's an positive integer now,
+ * but this may be changed in future, so it should be treated as an opaque value.
+ */
+type TimerHandle = number & { readonly '': unique symbol };
+
+/**
+ * Schedules the "immediate" execution of the given function after I/O events' callbacks.
+ *
```
drsm, line 618: Actually, after promise microtasks
```js
>> setImmediate(() => console.log('two')); require('fs').access('/', () => console.log('three')); Promise.resolve().then(() => console.log('one'));
Promise {
 
}
one
two
three
```

jirutka: I’ve copied this description from Node’s documentation. Microtask queue is also Node’s term. Does njs use the same terminology for this? @xeioex

xeioex: @jirutka  I would leave the description as is (we do not want to be too specific here, because it is an implementation detail).

>Does njs use the same terminology for this?

We generally prefer to use the same terminology to avoid confusion, but again, here, IMHO, it is too low level.

jirutka:
:information_source: Rebased on top of the current master to fix conflicts.

xeioex:
@jirutka  Committed, thanks.

PR #363, opened 2020-12-22
Fix off-by-3 in njs_string.c

disconnect3d:
TL;DR: The `base64url` string is of len 9 but we compared only its first 6 bytes.

This was found with a "cstrnfinder" research and I haven't tested this change (more info https://twitter.com/disconnect3d_pl/status/1339757359896408065). Close this PR if this change is incorrect.


lexborisov:
@disconnect3d 

Thanks for the patch!

PR #367, opened 2021-02-01
Fix dynamic code execution in safe mode

patryk4815:
Hi. I found a "Function" workaround and now in safe mode we can execute dynamic code like as "eval". 
I think it is security vuln and need CVE?

```
Function("){return 1337})//", "return this")();
```
BTW Why function arguments are not sanitized? It just concatenate string.

### DIFF
--- a/src/njs_function.c
+++ b/src/njs_function.c
@@ -898,6 +898,7 @@ njs_function_constructor(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
          */
 
         if (str.length != njs_length("return this")
+            || nargs != 1
             || memcmp(str.start, "return this", 11) != 0)
         {
             njs_type_error(vm, "function constructor is disabled"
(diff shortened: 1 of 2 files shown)


drsm:
@patryk4815 

nice catch!

but unlikely exploitable, 'cause in order to do this, the stuff before the function body should came from non-trusted source.

drsm:
BTW, it's would be nice to document that in "non-safe" mode it's up to the caller to sanitize inputs.

xeioex:
Hi @patryk4815,

Thanks for the patch, will be committed.

> I think it is security vuln and need CVE?

This is bug but not a security vulnerability, because in njs threat model the code is a part of nginx configuration, and the configuration is expected to be a trusted source (it also contains certificates, for example).

xeioex:
@patryk4815 

> BTW Why function arguments are not sanitized? It just concatenate string.

Mostly because dynamic execution methods like eval() and Function() are not expected to be used in njs at all.
Their support is limited, and was added to increase [test262](https://github.com/tc39/test262) coverage of other functionality.

Issue #251, opened 2019-11-19
Crypto add cipher and  decipher 


fishioon:
For aes cipher and decipher.

Sarniak1991:
Would be nice to see RSA verify in crypto :)

crasyangel:
@xeioex any process?

xeioex:
@crasyangel  we are in the process of adding node.js-like Buffer() object to work with binary data. After Buffer is done we plan to return to crypto features. Currently njs has byte-strings which is non-standard extension of ordinary String object.

xeioex:
@crasyangel, @drsm, @jirutka Feel free to test https://gist.github.com/xeioex/2d17327f13ca519750b4f5d6cf224e85

```
The following methods were implemented:
    crypto.getRandomValues()
    crypto.subtle.importKey()
        format: raw, pkcs8, spki
        algorithm: AES-CBC, AES-CTR, AES-GCM,
            ECDSA, HKDF, HMAC, PBKDF2,
            RSASSA-PKCS1-v1_5, RSA-OAEP, RSA-PSS
    crypto.subtle.decrypt()
    crypto.subtle.encrypt()
        algorithm: AES-CBC, AES-CTR, AES-GCM,
            RSA-OAEP
    crypto.subtle.deriveBits()
    crypto.subtle.deriveKey()
        algorithm: HKDF, PBKDF2
    crypto.subtle.digest()
        algorithm: SHA-1, SHA-256, SHA-384, SHA-512
    crypto.subtle.sign()
    crypto.subtle.verify()
        algorithm: ECDSA, HMAC, RSASSA-PKCS1-v1_5, RSA-PSS
```

see `test/webcrypto/` for examples

drsm:
@xeioex 

just found a [typo](https://gist.github.com/xeioex/2d17327f13ca519750b4f5d6cf224e85#file-webcrypto-patch-L2598).
everything else looks fine for me. thanks!

drsm:
BTW, it would be nice to test this against [WPT](https://github.com/web-platform-tests/wpt/tree/master/WebCryptoAPI), like they do it in [nodejs](https://github.com/nodejs/node/tree/master/test/wpt)

gallarda:
Will this WebCrypto API work with a Hardware Security Module (HSM)??

xeioex:
@xeioex 
>Will this WebCrypto API work with a Hardware Security Module (HSM)??

[...]

PR #414, opened 2021-08-14
Support for https URLs to the Fetch API.

abonavita:
Hello,

This is my first attempt at contributing code here, so please accept my apologies if I broke any of the rules. As the title said, this is a first try at https support for the Fetch API implemented in njs. Current implementation supports TLS name identification extension but comes with many limitations:
- It does not support server certificate verification (i.e. the traffic is subject to man-in-the-middle attacks).
- Protocols and ciphers advertised to server are respectively hard-coded to TLS1.0/1.1/1.2 and HIGH:!aNull:!MD5.
- TLS session caching cannot be disabled.

I am very eager to work on those limitations but I wanted first to:

1. Check that this is of interest to you.
2. Get your feedback on the code I wrote (so that I get the right habits early)
3. Discuss which of the limitations really need to be addressed and how. My personal opinion is that configurable verification is a "must", configurable protocols and ciphers are "nice to have" and session caching is really not worth it.

Thanks for your feedback,

A.

### DIFF
--- a/nginx/config
+++ b/nginx/config
@@ -19,6 +19,25 @@ if [ $HTTP != NO ]; then
     fi
 fi
 
+if [ $USE_OPENSSL != NO ]; then
+    OPENSSLDIR=`openssl version -d | cut -d '"' -f 2`
+    if [ -d "$OPENSSLDIR" ]; then
+        if [ -f "$OPENSSLDIR/cert.pem" ]; then
+            have=NGX_OPENSSL_CERT_FILE value="\"$OPENSSLDIR/cert.pem\""
+            . auto/define
+        elif [ -d "$OPENSSLDIR/certs" ]; then
+            have=NGX_OPENSSL_CERT_DIR value="\"$OPENSSLDIR/certs\""
+            . auto/define
+        else
+            echo "Cannot find openssl truststore (openssldir=$OPENSSLDIR)"
+            exit 1
+        fi
+    else
+        echo Cannot find openssldir
+        exit 1
+    fi
+fi
+
 if [ $STREAM != NO ]; then
     ngx_module_type=STREAM
     ngx_module_name=ngx_stream_js_module
--- a/nginx/ngx_js.c
+++ b/nginx/ngx_js.c
@@ -110,6 +110,20 @@ ngx_js_integer(njs_vm_t *vm, njs_value_t *value, ngx_int_t *n)
 
 
 ngx_int_t
+ngx_js_boolean(njs_vm_t *vm, njs_value_t *value, njs_bool_t *b)
+{
+    if (!njs_value_is_boolean(value)) {
+        njs_vm_error(vm, "is not a boolean");
+        return NGX_ERROR;
+    }
+
+    *b = njs_value_bool(value);
+
+    return NGX_OK;
+}
+
+
+ngx_int_t
 ngx_js_string(njs_vm_t *vm, njs_value_t *value, njs_str_t *str)
 {
     if (value != NULL && !njs_value_is_null_or_undefined(value)) {
--- a/nginx/ngx_js.c
+++ b/nginx/ngx_js.c
@@ -110,20 +110,6 @@ ngx_js_integer(njs_vm_t *vm, njs_value_t *value, ngx_int_t *n)
 
 
 ngx_int_t
-ngx_js_boolean(njs_vm_t *vm, njs_value_t *value, njs_bool_t *b)
-{
-    if (!njs_value_is_boolean(value)) {
-        njs_vm_error(vm, "is not a boolean");
-        return NGX_ERROR;
-    }
-
-    *b = njs_value_bool(value);
-
-    return NGX_OK;
-}
-
-
-ngx_int_t
 ngx_js_string(njs_vm_t *vm, njs_value_t *value, njs_str_t *str)
 {
     if (value != NULL && !njs_value_is_null_or_undefined(value)) {
--- a/nginx/ngx_js.h
+++ b/nginx/ngx_js.h
@@ -65,6 +65,7 @@ ngx_int_t ngx_js_core_init(njs_vm_t *vm, ngx_log_t *log);
 
 ngx_int_t ngx_js_string(njs_vm_t *vm, njs_value_t *value, njs_str_t *str);
 ngx_int_t ngx_js_integer(njs_vm_t *vm, njs_value_t *value, ngx_int_t *n);
+ngx_int_t ngx_js_boolean(njs_vm_t *vm, njs_value_t *value, njs_bool_t *b);
 
 
 #endif /* _NGX_JS_H_INCLUDED_ */
--- a/nginx/ngx_js.h
+++ b/nginx/ngx_js.h
@@ -65,7 +65,6 @@ ngx_int_t ngx_js_core_init(njs_vm_t *vm, ngx_log_t *log);
 
 ngx_int_t ngx_js_string(njs_vm_t *vm, njs_value_t *value, njs_str_t *str);
 ngx_int_t ngx_js_integer(njs_vm_t *vm, njs_value_t *value, ngx_int_t *n);
-ngx_int_t ngx_js_boolean(njs_vm_t *vm, njs_value_t *value, njs_bool_t *b);
 
 
 #endif /* _NGX_JS_H_INCLUDED_ */
--- a/nginx/ngx_js_fetch.c
+++ b/nginx/ngx_js_fetch.c
@@ -57,9 +57,9 @@ struct ngx_js_http_s {
     in_port_t                      port;
 
 #if (NGX_SSL)
-    ngx_int_t                      ssl_verify;
     ngx_ssl_t                     *ssl;
     ngx_str_t                      tls_name;
+    njs_bool_t                     ssl_verify;
 #endif
 
     ngx_peer_connection_t          peer;
@@ -486,11 +486,8 @@ ngx_js_ext_fetch(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
         if (is_https) {
 
             value = njs_vm_object_prop(vm, init, &verify_key, &lvalue);
-            if (value != NULL
-                && ngx_js_integer(vm, value, &http->ssl_verify) != NGX_OK)
-            {
-                njs_vm_error(vm, "verify is not an integer");
-                goto fail;
+            if (value != NULL) {
+                http->ssl_verify = njs_value_bool(value);
             }
 
             value = njs_vm_object_prop(vm, init, &trusted_key, &lvalue);
--- a/nginx/ngx_js_fetch.c
+++ b/nginx/ngx_js_fetch.c
@@ -10,6 +10,7 @@
 #include <ngx_core.h>
 #include <ngx_event.h>
 #include <ngx_event_connect.h>
+#include <njs_unix.h>
 #include "ngx_js.h"
 
 
@@ -59,6 +60,7 @@ struct ngx_js_http_s {
 #if (NGX_SSL)
     ngx_ssl_t                     *ssl;
     ngx_str_t                      tls_name;
+    njs_bool_t                     ssl_verify;
 #endif
 
     ngx_peer_connection_t          peer;
@@ -151,8 +153,11 @@ static njs_int_t ngx_response_js_ext_body(njs_vm_t *vm, njs_value_t *args,
 static void ngx_js_http_ssl_init_connection(ngx_js_http_t *http);
 static void ngx_js_http_ssl_handshake_handler(ngx_connection_t *c);
 static void ngx_js_http_ssl_handshake(ngx_js_http_t *http);
-static njs_int_t ngx_js_http_set_ssl(ngx_js_http_t *http);
+static njs_int_t ngx_js_http_set_ssl(ngx_js_http_t *http,
+    const char *truststore, ngx_flag_t truststore_is_file,
+    ngx_int_t verify_depth);
 static njs_int_t ngx_js_http_ssl_name(ngx_js_http_t *http);
+static int ngx_js_http_ssl_verify_callback(int ok, X509_STORE_CTX *x509_store);
 #endif
 
 static njs_external_t  ngx_js_ext_http_response_headers[] = {
@@ -350,12 +355,26 @@ ngx_js_ext_fetch(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
     ngx_resolver_ctx_t  *ctx;
     njs_external_ptr_t   external;
     njs_opaque_value_t  *start, lvalue, headers_value;
+#if (NGX_SSL)
+    char                *p, *truststore;
+    ngx_int_t            depth;
+    njs_str_t            truststore_param;
+    ngx_flag_t           is_https, truststore_is_file;
+    char                 buf[NJS_MAX_PATH + 1];
+#endif
 
     static const njs_str_t body_key = njs_str("body");
     static const njs_str_t headers_key = njs_str("headers");
     static const njs_str_t buffer_size_key = njs_str("buffer_size");
     static const njs_str_t body_size_key = njs_str("max_response_body_size");
     static const njs_str_t method_key = njs_str("method");
+#if (NGX_SSL)
+    static const njs_str_t verify_key = njs_str("verify");
+    static const njs_str_t verify_depth_key = njs_str("verify_depth");
+    static const njs_str_t trusted_key = njs_str("trusted_certificate");
+
+    is_https = 0;
+#endif
 
     external = njs_vm_external(vm, NJS_PROTO_ID_ANY, njs_argument(args, 0));
     if (external == NULL) {
@@ -403,10 +422,16 @@ ngx_js_ext_fetch(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
         u.url.len -= 8;
         u.url.data += 8;
         u.default_port = 443;
-        if (ngx_js_http_set_ssl(http) != NJS_OK) {
-            ngx_js_http_error(http, 0, "js http failed to create ssl object");
-            goto fail;
-        }
+        is_https = 1;
+        http->ssl_verify = 1;
+#ifdef NGX_OPENSSL_CERT_FILE
+        truststore_is_file = 1;
+        truststore = NGX_OPENSSL_CERT_FILE;
+#else
+        truststore_is_file = 0;
+        truststore = NGX_OPENSSL_CERT_DIR;
+#endif
+        depth = -1;
 #endif
 
     } else {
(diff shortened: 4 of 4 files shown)


xeioex:
Hi @abonavita,

Thanks for the contribution. I glad to see a new contributor.

> Check that this is of interest to you.

yes, we are interested in supporting HTTPS protocol in fetch API.

> Get your feedback on the code I wrote (so that I get the right habits early)

At the first glance the code looks good, I will take a look the next week deeper.

> My personal opinion is that configurable verification is a "must", configurable protocols and ciphers are "nice to have" and session caching is really not worth it.

I agree with your analysis.

xeioex:
Hi @crasyangel, @jirutka , @patryk4815, @wxiaoguang,  

Feel free to test the HTTPS fetch API support.

abonavita:
@xeioex ,

Thanks for the quick feedback. regarding the verification, I was thinking of an extra "verify" parameter in the options that could be:
- false : no verification at all. This is the current behavior and to be in line with proxy_ssl_verify we could make it the default.
- An object with two mandatory fields : "trusted_certificate" that contains the path to a PEM file and "depth" the indicates the maximum depth of verification for the certificates chain.

Let me know what you think.

xeioex:
> @xeioex ,
> 
> Thanks for the quick feedback. regarding the verification, I was thinking of an extra "verify" parameter in the options that could be:
> 
> * false : no verification at all. This is the current behavior and to be in line with proxy_ssl_verify we could make it the default.
> * An object with two mandatory fields : "trusted_certificate" that contains the path to a PEM file and "depth" the indicates the maximum depth of verification for the certificates chain.
> 
> Let me know what you think.

Looks good.

xeioex:
@abonavita 

Please, also do not forget to add/modify [nginx tests](https://github.com/nginx/nginx-tests).

The way I run them is as follows (from nginx-tests repo):
`rm -fr /tmp/nginx-test* && TEST_NGINX_LEAVE=1 TEST_NGINX_BINARY=/nginx/binary/path/objs/nginx prove js*.t`

You may add additional test into `js_fetch.t`.

xeioex:
@abonavita 

>At the first glance the code looks good, I will take a look the next week deeper.

The code overall is good, I will apply some minor style changes before the commit.

abonavita:
@xeioex,
> The way I run them is as follows (from nginx-tests repo):
> `rm -fr /tmp/nginx-test* && TEST_NGINX_LEAVE=1 TEST_NGINX_BINARY=/nginx/binary/path/objs/nginx prove js*.t`
> 
> You may add additional test into `js_fetch.t`.
Started looking into this. I can easily add a test for https. But I have a problem if I want to test TLS extension. For it to work properly, I need a DNS resolver to resolve to 127.0.0.1 the URLs I want to test. I can think of 2 approaches : 
- Use a public DNS (8.8.8.8) and entries like lacolhost.com and localtest.me
- Implement a "fake DNS" using Net::DNS::Nameserver (that adds an extra dependency).
Any preference ? Or should I drop the idea of including an automated test for this here ?

> The code overall is good, I will apply some minor style changes before the commit.
Thanks,

xeioex:
@abonavita  resolver related tests are here https://github.com/nginx/nginx-tests/blob/master/js_fetch_resolver.t

abonavita:
@xeioex 

Thanks for the pointer to the js_fetch_resolver.t.
Submitted a merge request for nginx-test [here](https://github.com/nginx/nginx-tests/pull/3)

xeioex:
Hi @abonavita,

>It does not support server certificate verification (i.e. the traffic is subject to man-in-the-middle attacks).

BTW, do you plan to update the patch with the verify feature?

abonavita:
Hi @xeioex,

Yes, I do plan to update the patches (for njs and nginx-tests) with the verify feature (hopefully during the week-end or early next week).
My main concern right now is about the default value : 
- Should we make it false by default (much like the proxy_ssl_verify) ?  I don't really like it because users can create something completely insecure without even realizing they do.
- Should we make the parameter mandatory ? I don't like this one because it does not "just work out of the box".
If you have other ideas, I'm very open to suggestions...

xeioex:
@abonavita 


> * Should we make it false by default (much like the proxy_ssl_verify) ?  I don't really like it because users can create something completely insecure without even realizing they do.

I agree with you. `proxy_ssl_verify` is `off` by default only for historical reasons, and this behaviour is not acceptable for the fetch API in 2021.

> * Should we make the parameter mandatory ? I don't like this one because it does not "just work out of the box".
>   If you have other ideas, I'm very open to suggestions...

I believe the verify parameter should be `on` by default, but it should be possible to disable it.

>it does not "just work out of the box".

Also, the setup should be relatively easy. The default value for "trusted_certificate" should be a system's location of the CA certificate(s). For example, for Ubuntu it is `/etc/ssl/certs/ca-certificates.crt`.

Additional links:

Nginx uses [SSL_CTX_load_verify_locations](https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_load_verify_locations.html) which accepts a path to a CA bundle on the system.

[node.js way](https://nodejs.org/api/cli.html#cli_use_bundled_ca_use_openssl_ca):
```
  --use-bundled-ca, --use-openssl-ca
             Use bundled Mozilla CA store as supplied by current Node.js version or use OpenSSL's default CA store.  The de‐
             fault store is selectable at build-time.

             The bundled CA store, as supplied by Node.js, is a snapshot of Mozilla CA store that is fixed at release time.  It
             is identical on all supported platforms.

             Using OpenSSL store allows for external modifications of the store.  For most Linux and BSD distributions, this
             store is maintained by the distribution maintainers and system administrators.  OpenSSL CA store location is de‐
             pendent on configuration of the OpenSSL library but this can be altered at runtime using environment variables.

             See SSL_CERT_DIR and SSL_CERT_FILE.
```

Ideally we want to set the default value for "trusted_certificate" to OPENSSL CA certs location at build-time.

abonavita:
@xeioex,

OK. Just to make sure I got this right, here are a few commented examples :
`var url = 'https://nginx.org';
ngx.fetch(url); // Checks server certificate using OPENSSL CA certs location determined at build time and a default depth of 1 (same default as proxy_ssl_verify_depth - seems pretty low to me).
ngx.fetch(url, {verify:false}); // Does not check server certificate - subject to MIM attacks
ngx.fetch(url, {trusted_certificate: '/root/myca-bundle.crt'}); // Uses the specified location instead of default.
ngx.fetch(url, {verify_depth: 3}); // In case the certificate chain is pretty long and the default value is not enough.`

Please correct me if I misunderstood anything.

xeioex:
@abonavita 

>Please correct me if I misunderstood anything.

everything is correct. I also agree the default depth should be higher than 1.

abonavita:
@xeioex 

Quoting from [SSL_CTX_set_verify](https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_verify.html) :
> The default depth limit is 100, allowing for the peer certificate and additional 100 CA certificates.

So, we should not try to set any default if none was provided by the user and let OpenSSL use its own default. What do you think ?

xeioex:
@abonavita 

>OpenSSL use its own default. What do you think ?

LGTM.

abonavita:
@xeioex,

Pushed an update (here and on nginx-test) with the options discussed. I had some issues with veirfy_depth (which behavior apparently changed with openSSL1.1.0) but it should be OK now.

Let me know if you need me to change anything.

xeioex:
@abonavita Great work.

I merged the patches into a single [patch](https://gist.github.com/43392c1c976641710cbaa41459ef01d8). 
I would prefer to review a single patch later (the same is for tests). 

The bool patch is removed, as in JS arguments are typically coerced and not strictly type-checked.

remaining issues with the patch that I found so far

1) openssl objects are not destroyed and not freed

```
Direct leak of 6280 byte(s) in 1 object(s) allocated from:
    #0 0x7f2961dddbc8 in malloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dbc8)
    #1 0x7f29618ba94d in CRYPTO_zalloc (/lib/x86_64-linux-gnu/libcrypto.so.1.1+0x17b94d)
    #2 0x55795b4d0a85 in ngx_js_http_ssl_init_connection nginScript/njs/nginx/ngx_js_fetch.c:914
    #3 0x55795b4d1b31 in ngx_js_http_write_handler nginScript/njs/nginx/ngx_js_fetch.c:1121
    #4 0x55795b1f744e in ngx_epoll_process_events src/event/modules/ngx_epoll_module.c:930
    #5 0x55795b1bd308 in ngx_process_events_and_timers src/event/ngx_event.c:247
    #6 0x55795b1ea5da in ngx_single_process_cycle src/os/unix/ngx_process_cycle.c:300
    #7 0x55795b12e9b1 in main src/core/nginx.c:380
    #8 0x7f2960be60b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
```

to catch memory leaks use [Address Sanitizer ](http://nginx.org/en/docs/dev/development_guide.html#debug_memory
)
I guess we need here something similar to [ngx_http_close_connection](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_request.c#L3771) 
instead of ngx_close_connection().
 
2) in js_fetch_https.t please revert '\t' back. 
The coding style for perls tests: 
we use tabs to indent the perl code, but we use spaces in nginx.conf

3) I am not also sure that we need `NGX_OPENSSL_CERT_DIR` as nginx does not support it.
4) `#include <njs_unix.h>` is not needed, as there is `NGX_MAX_PATH`

abonavita:
Thanks @xeioex .

So, you want me to provide two HG changesets (one for njs and one for nginx-tests) ? I can definitely do that.

Regarding NGX_OPENSSL_CERT_DIR, I went this route because openssl first looks into its openssldir/cert.pem, then uses openssl/certs directory (source https://www.madboa.com/geek/openssl/#what-certificate-authorities-does-openssl-recognize) and my debian setup shows no cert.pem file but a certs directory. So, on Debian the expected default is de facto a directory. It appears that this directory includes a certificate with all the certificates concatenated but it's not where openssl expects it (and there is no link). If you think we should go another route, just let me know.

For the rest of your comments "will do".

xeioex:
@abonavita 

>So, you want me to provide two HG changesets (one for njs and one for nginx-tests) ? I can definitely do that.

that would be great, thanks!

>Regarding NGX_OPENSSL_CERT_DIR, 

agree, it is better to work out of the box on such systems. So, leave it as is.

abonavita:
@xeioex 

> The bool patch is removed, as in JS arguments are typically coerced and not strictly type-checked.
If I use `[ngx_js_integer](https://github.com/nginx/njs/blob/master/nginx/ngx_js.c#L99)`, then the following javascript code does not work : 
```javascript
ngx.fetch("https://nginx.org", {verify: false});
```
Of course, this works:
```javascript
ngx.fetch("https://nginx.org", {verify: 1});
```
I could change [ngx_js_integer](https://github.com/nginx/njs/blob/master/nginx/ngx_js.c#L99) to return `0` on `false` and `1` on `true` but I don't think this is what you had in mind.
So, the only option I see left is to change the API and document verify as an integer.
Let me know what you think.

xeioex:
@abonavita 

I suggested to use the following API
```c
    value = njs_vm_object_prop(vm, init, &verify_key, &lvalue);
    if (value != NULL) {
        http->ssl_verify = njs_value_bool(value);
    }
```

abonavita:
@xeioex,

Here are the two changesets : [https.njs.patch](https://gist.github.com/abonavita/3c22606ac63a6bb1e077237875eb1739).

Thanks for the help with the boolean. There is a lot I don't know about the njs code base and API...

Also, the use of no_wait_shutdown when shutting down SSL connexion is inspired from http upstream. If not present, there is a "use-after-free" error when handshake does not go through because the root certificate was not found.

Comments always welcome.

abonavita:
@xeioex,

Started working on this and I am wondering if it wouldn't be better to keep verify (the boolean) as a parameter and depth+trusted certificates as directives. This way, you can have different strategies in your code (e.g. verify on for external calls and off for internal ones) and keep the creation of the ngx_ssl_t object at configuration time.

Let me know what you think.

xeioex:
@abonavita 

>Let me know what you think.

agree.

abonavita:
Hi @xeioex ,

Here are two new changesets for your review : [njs-fetch-https-1sep21](https://gist.github.com/abonavita/d3a6ef4e0a2822b287111272c7461dbf).

Comments always welcome.

xeioex:
@abonavita 

>Here are two new changesets for your review : njs-fetch-https-1sep21.
>
>Comments always welcome.

Thanks, I am on it.

xeioex:
@abonavita 

While reviewing your latest patch I found out that TLS session caching is not working (and is not going to work easily).

Nginx uses OpenSSL API [SSL_CTX_set_session_cache_mode](https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set_session_cache_mode.html) with `SSL_SESS_CACHE_CLIENT` which is not work out of the box for TLS clients. For caching to work we need to store the SSL session somewhere, not linking it to the current request. Later using [SSL_set_session](https://www.openssl.org/docs/man1.1.0/man3/SSL_set_session.html) to manually set existing SSL session for a known peer. 

Nginx uses [ngx_http_upstream_rr_peer_t](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_upstream_round_robin.h#L44) structure store the SSL sessions, it store sessions in a [shared memory](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_upstream_round_robin.c#L676).

As a result, at this point, I think we should postpone the caching feature.

Here is 2 changes I am going to apply to your recent patch (feel free to report any issues)
1) some minor changes, mostly style https://gist.github.com/xeioex/1c180c96163541c4ad700c6c2d7a802c
2) removing TLS caching attempts https://gist.github.com/xeioex/48962dd065e243a53923d10a56d6e6fd

The last question to solve before the commit are: 
* the default value for `js_fetch_trusted_certificate`
* SSL_CTX_load_verify_locations() directory usage 

I am still thinking about it.

abonavita:
@xeioex ,


> While reviewing your latest patch I found out that TLS session caching is not working (and is not going to work easily).
Thanks for figuring that out. I was miles away from suspecting this.

> As a result, at this point, I think we should postpone the caching feature.
100% with you on this one. Actually, I was thinking about HTTP/1.1 support for this API (which, if I am correct, brings similar issues) and realized that was probably huge.

> 1. some minor changes, mostly style https://gist.github.com/xeioex/1c180c96163541c4ad700c6c2d7a802c
> 2. removing TLS caching attempts https://gist.github.com/xeioex/48962dd065e243a53923d10a56d6e6fd
Both look good to me. And thanks for figuring out how to get rid of is_https (tried a couple of times and failed because I did not dare set ssl_verify for http requests).

> The last question to solve before the commit are:
> 
> * the default value for `js_fetch_trusted_certificate`
> * SSL_CTX_load_verify_locations() directory usage
Yeah, I don't find the solution elegant as it stands now, but I cannot get myself to drop the "works out of the box on Debian" requirement.

Thanks a lot for you help walking me through all this and your patience.

xeioex:
@abonavita 

FYI, I plan to finalise the patch next week. (Doing https://github.com/nginx/njs/issues/251 right now).

crasyangel:
@abonavita @xeioex 
1. why not use ngx_ssl_trusted_certificate simply?
2. http->ssl do not need ngx_external_ssl.
`   r = c->data;
   conf = ngx_http_get_module_loc_conf(r, ngx_http_js_module);
   http->ssl = conf->ssl;`
3. js_ssl_protocols, js_ssl_ciphers, js_ssl_crl, js_ssl_conf_command should be Introduced

crasyangel:
two more questions:
1, ngx_js_http_ssl_init_connection must have a return value. And should first called when ngx_js_http_connect.
ensure call ngx_js_fetch_result before ngx_js_http_fetch_done, or call ngx_js_fetch_result with a error
2, NGX_OPENSSL_CERT_FILE macro is not good. Default is unnecessary, such as proxy_ssl_trusted_certificate

xeioex:
@crasyangel 

Thanks for your feedback.

>why not use ngx_ssl_trusted_certificate simply?
>NGX_OPENSSL_CERT_FILE macro is not good.  Default is unnecessary, such as proxy_ssl_trusted_certificate

yes, I came to the same conclusion. ngx_ssl_trusted_certificate() will be used, and all default macros will be removed.

> http->ssl do not need ngx_external_ssl.
>r = c->data; conf = ngx_http_get_module_loc_conf(r, ngx_http_js_module); http->ssl = conf->ssl;

Because ngx_js_fetch.c is a generic code, which should work for [http](http://nginx.org/en/docs/http/ngx_http_js_module.html) as well as for [stream](http://nginx.org/en/docs/stream/ngx_stream_js_module.html) module.

> js_ssl_protocols, js_ssl_ciphers, js_ssl_crl, js_ssl_conf_command should be Introduced

Yes, they make sense, but can be added later.

> 1, ngx_js_http_ssl_init_connection must have a return value. And should first called when ngx_js_http_connect.
ensure call ngx_js_fetch_result before ngx_js_http_fetch_done, or call ngx_js_fetch_result with a error

Yes, I was thinking about this place.

abonavita:
@crasyangel and @xeioex ,

Thank you both for shedding new light on this.

> > why not use ngx_ssl_trusted_certificate simply?
> > NGX_OPENSSL_CERT_FILE macro is not good.  Default is unnecessary, such as proxy_ssl_trusted_certificate
>
> yes, I came to the same conclusion. ngx_ssl_trusted_certificate() will be used, and all default macros will be removed.

By default proxy_ssl_trusted_certificate is not set, but proxy_ssl_verify is off. As a consequence, it "just works" out of the box (but could be considered "insecure", YMMV).
With the suggested approach (except if I missed something and please correct me if so), fetching from https won't work "out of the box" (verify is on by default and should be).
Using the certificate configured with ssl_trusted_certificate (much like OCSP does) has the advantage of simplicity (no new directive to introduce) but introduces a dependency to the http_ssl module. Realistically, I think it's fine (I have a hard time imagining a use case where a server wants to fetch https ressources without supporting https itself). If so, I will use the NGX_HTTP_SSL macro to guard the https code instead of NGX_SSL.

If we all agree that "it works out of the box" is not a requirement for this feature, I definitely agree with you.

Let me know what you think.
 
> > 1, ngx_js_http_ssl_init_connection must have a return value. And should first called when ngx_js_http_connect.
> > ensure call ngx_js_fetch_result before ngx_js_http_fetch_done, or call ngx_js_fetch_result with a error
> 
> Yes, I was thinking about this place.

I think it makes sense. It's just my lack of familiarity with the code base that got me walking this route.

I might have a hard time working on this next week (will try though), but the week after should be fine. Will post an updated patch here.

xeioex:
@abonavita  proxy_ssl_verify will be on by default

abonavita:
@xeioex 

> I also plan to finalize your patch be myself, so no additional input from you is required.

Great. Just a side note I noticed when applying your latest changes : you are missing a `pscf->ssl = ssl;` around
https://gist.github.com/xeioex/48962dd065e243a53923d10a56d6e6fd#file-https_ssl-patch-L125.

Without it, the fetch in stream talk "plain HTTP" to the https server...

Hope this helps,

thresheek:

xeioex:
@abonavita, @crasyangel  Thanks you for your contribution. [Committed](https://github.com/nginx/njs/commit/c98ed0f1d2a84c36dccd4f4ddf0cea8906bd24e5).

PR #605, opened 2022-12-15
Typo in error message

kapouer:


kapouer:
Worth noting that I refrained from correcting all "SUBSTRACT*" since it would affect the VM code.

xeioex:
Thanks, committed in https://github.com/nginx/njs/commit/f6c64448f6e0a4fab544e37fa6e0c8dcf7142c6a.

PR #626, opened 2023-03-22
Fix compiler detection when CC has a wrapper

orgads:
For example, CC='ccache gcc'


xeioex:
Committed in https://github.com/nginx/njs/commit/247daa049057db583a2ab50a300bb4ee13e3e746, thanks.

PR #628, opened 2023-03-22
Fix uninitialized variable in math

orgads:


xeioex:
Hi @orgads,

How can I reproduce the issue? What compiler do you use?

orgads:
Very old GCC 4.1.2 :)

xeioex:
Hi @orgads,

Unlike nginx, in njs we are not trying to be compatible with any C compiler. 
In nginx there is NGX_SUPPRESS_WARN macro, which allows to put overhead initialization code for compilers which fail to see that initialization is unnecessary.  We do not have such macro in njs.

We are trying to be compatible with modern (available on active and working systems) compilers on UNIX-like systems though.

If you nevertheless want to compile with gcc 4.1 consider providing extra compilation options using `--cc-opt="-w"` to suppress the warnings.

orgads:
Is there a way to pass configure flags when building as a module?

xeioex:
@orgads,

njs build [inherits](https://github.com/nginx/njs/blob/master/nginx/config.make#L6) the CC and CFLAGS from nginx build.

orgads:
I see, thank you.

xeioex:
@orgads 

Just provide `--with-cc-opt='-w'` to nginx configure.

PR #629, opened 2023-03-22
RFC: Support compilation on Windows

orgads:
Some features are stubs, but at least it compiles ;)


xeioex:
Hi @orgads,

Thanks for contribution, but we have no plans for Windows support.
Because it is not only requires modifying code, but also maintenance.

orgads:
Yeah, this was an RFC in the first place. I'll just keep it locally. Closing.

PR #639, opened 2023-05-12
Always use a sharp (#) symbol as a sed(1) separator.

osokin:
sed(1) command line utility may fail with the following error:
sed: 1: "s, at EXTRA_LIBS@,-lm   -L ...": bad in substitute command: '-'
when a replacement for @EXTRA_LIBS@ contains a comma symbol.


xeioex:
Committed in https://github.com/nginx/njs/commit/e9d814a0a8c1857f78fdc85ed1461cb9101ebf4e.

PR #649, opened 2023-06-10
Types: added TS types for TextDecoder and TextEncoder.

jirutka:
/cc @xeioex


xeioex:
Committed in https://github.com/nginx/njs/commit/8cdf3cd6564cd64d61c8bc45c4fa14c896ba6212, thanks.

PR #715, opened 2024-05-28
Test262: improved skipping of tests.

xeioex:
The skipped tests (which are expected to fail) are not executed in order to make output clearer.


VadimZhestikov (approved):
Looks good

PR #716, opened 2024-05-28
Fixed RegExp.prototype[@@replace]().

xeioex:
With replacement containing "$'", "$\`" and Unicode characters.

The similar issue was fixed for String.prototype.replace() in 10127d70e941 (0.7.11).

Found by OSS-Fuzz.


VadimZhestikov (approved):
Looks good

PR #726, opened 2024-06-02
Modules: shared dictionary add, set, incr methods timeout argument.

jo-carter:
The optional timeout argument overrides the timeout specified with
the shared_dict_zone directive for the effected key and operation
only. The timeout is specified in milliseconds.

This is useful when the majority of keys are expected to require
unique timeouts.

### Proposed changes

Describe the use case and detail of the change. If this PR addresses an issue on GitHub, make sure to include a link to that issue using one of the [supported keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) here in this description (not in the title of the PR).

### Checklist

### DIFF
--- a/nginx/ngx_js_shared_dict.c
+++ b/nginx/ngx_js_shared_dict.c
@@ -85,15 +85,16 @@ static ngx_js_dict_node_t *ngx_js_dict_lookup(ngx_js_dict_t *dict,
 #define NGX_JS_DICT_FLAG_MUST_NOT_EXIST   2
 
 static ngx_int_t ngx_js_dict_set(njs_vm_t *vm, ngx_js_dict_t *dict,
-    njs_str_t *key, njs_value_t *value, unsigned flags);
+    njs_str_t *key, njs_value_t *value, ngx_msec_t timeout, unsigned flags);
 static ngx_int_t ngx_js_dict_add(ngx_js_dict_t *dict, njs_str_t *key,
-    njs_value_t *value, ngx_msec_t now);
+    njs_value_t *value, ngx_msec_t timeout, ngx_msec_t now);
 static ngx_int_t ngx_js_dict_update(ngx_js_dict_t *dict,
-    ngx_js_dict_node_t *node, njs_value_t *value, ngx_msec_t now);
+    ngx_js_dict_node_t *node, njs_value_t *value, ngx_msec_t timeout,
+    ngx_msec_t now);
 static ngx_int_t ngx_js_dict_get(njs_vm_t *vm, ngx_js_dict_t *dict,
     njs_str_t *key, njs_value_t *retval);
 static ngx_int_t ngx_js_dict_incr(ngx_js_dict_t *dict, njs_str_t *key,
-    njs_value_t *delta, njs_value_t *init, double *value);
+    njs_value_t *delta, njs_value_t *init, double *value, ngx_msec_t timeout);
 static ngx_int_t ngx_js_dict_delete(njs_vm_t *vm, ngx_js_dict_t *dict,
     njs_str_t *key, njs_value_t *retval);
 static ngx_int_t ngx_js_dict_copy_value_locked(njs_vm_t *vm,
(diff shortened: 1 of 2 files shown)


jo-carter:
Keyval_zone has the same feature for updates over the N+ API, for reference: https://nginx.org/en/docs/http/ngx_http_api_module.html#def_nginx_http_keyval_zone_post_patch

nginx/ngx_js_shared_dict.c
```diff
@@ -765,7 +767,32 @@ njs_js_ext_shared_dict_incr(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
         njs_value_number_set(init, 0);
     }
 
-    rc = ngx_js_dict_incr(shm_zone->data, &key, delta, init, &value);
+    timeout_src = njs_arg(args, nargs, 4);
```
xeioex, line 770: Style: I suggest to use a shorter variable name here.

nginx/ngx_js_shared_dict.c
```diff
@@ -765,7 +767,32 @@ njs_js_ext_shared_dict_incr(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
...
+    if (!njs_value_is_undefined(timeout_src)) {
+
```
xeioex, line 772: Style: excessive empty line.

nginx/ngx_js_shared_dict.c
```diff
@@ -765,7 +767,32 @@ njs_js_ext_shared_dict_incr(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
...
+        timeout = (ngx_msec_t) njs_value_number(timeout_src);
+
+        if (timeout < 1) {
+            njs_vm_type_error(vm,
+                              "timeout must be greater than or equal to 1");
+            return NJS_ERROR;
+        }
+
+        if (!dict->timeout) {
```
xeioex, line 786: Probably we should check that before timeout extraction.

nginx/ngx_js_shared_dict.c
```diff
@@ -967,7 +995,32 @@ njs_js_ext_shared_dict_set(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
         }
     }
 
-    rc = ngx_js_dict_set(vm, shm_zone->data, &key, value, flags);
+    timeout_src = njs_arg(args, nargs, 3);
+    if (!njs_value_is_undefined(timeout_src)) {
+
```
xeioex, line 1000: Style: excessive empty line.

nginx/ngx_js_shared_dict.c
```diff
@@ -967,7 +995,32 @@ njs_js_ext_shared_dict_set(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
         }
     }
 
-    rc = ngx_js_dict_set(vm, shm_zone->data, &key, value, flags);
+    timeout_src = njs_arg(args, nargs, 3);
```
xeioex, line 998: Style: I suggest to use a shorter variable name here.

xeioex (changes requested):
Hi @jo-carter, good patch in general.
I have some minor improvements. We also need auto tests here.

1) I believe we should remove Example: part from the commit log.
2) Instead we need an auto test.

See minor improvements and the auto tests here: https://gist.github.com/xeioex/dcfebd3512aa56fd180f6b243e79d071.

Feel free to modify or improve auto tests.
I run them as follows from njs directory:
```TEST_NGINX_LEAVE=1  TEST_NGINX_BINARY=/path/to/nginx/binary/nginx prove -r -I /path/to/nginx-tests/lib/ nginx/t/js_shared_dict.t```

nginx/ngx_js_shared_dict.c
jo-carter: Agreed, makes sense.

jo-carter:
@xeioex Thank you for the review and patches provided. Agreed on all points regarding style and formatting, and tests.

I made two adjustments to the patches you provided:
1. For consistency I adjusted the shortened variable name for njs_value_t for timeout argument to match in both _njs_js_ext_shared_dict_set_ and _njs_js_ext_shared_dict_incr_ - both are now timeo. 
2. I've re-arranged order of new tests to make false outcome less likely(similar to "bar keys after a delay" test). Probably an explicit sleep would be better, however ms level sleeps seem to require importing non-core modules in perl.

xeioex (approved):

jo-carter:
@xeioex Thanks again :)

PR #735, opened 2024-06-12
HTTP: fixed r.subrequest() error handling.

xeioex:
Previously, when at least 2 subrequests were scheduled they both succeed, but the callback for the second threw an exception heap-use-after-free happened.

The issue was introduced in 0.8.1 (4cb8e873e8c6).

### Checklist


p-pautov:
May be elaborate a bit on the reasons behind use-after-free? For example, "there was a nested chain of ngx_http_run_posted_requests() calls and terminating request in the inner call leaved outer calls with already freed request pointer".

p-pautov (approved):

PR #754, opened 2024-07-02
Fixed Function constructor handling when called without arguments.

VadimZhestikov:
### Proposed changes

Fixed Function w/o args called relative an object.

### Checklist

- [X] I have read the [`CONTRIBUTING`](CONTRIBUTING.md) document
- [X] If applicable, I have added tests that prove my fix is effective or that my feature works
- [X] If applicable, I have checked that any relevant tests pass after adding my changes

### DIFF
--- a/src/njs_function.c
+++ b/src/njs_function.c
@@ -1054,9 +1054,11 @@ njs_function_constructor(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
 
     njs_chb_append_literal(&chain, "){");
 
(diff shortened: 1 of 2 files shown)


xeioex (changes requested):
I suggest to improve the commit message, something like

```
Fixed Function constructor handling when called with zero arguments.

Corrected the behavior of Function.constructor() when invoked without arguments relative to an object.
```

xeioex:
The commit log message length being less than 67 chars still applies.

xeioex (changes requested):
The commit log message length being less than 67 chars still applies.

xeioex (approved):

PR #768, opened 2024-07-30
fix(types): NgxKeyValuePair is actually an array

TPXP:
### Proposed changes

Correct the type definition for `NgxSharedDict.items()`. At runtime, I found out it returns an array of `[key, value]` objects, just like `Object.items()`. As documented here https://nginx.org/en/docs/njs/reference.html#dict_items

### Checklist

## Minimal reproduction
```
function dumpItems(r) {
  ngx.shared.myarray.set("hello", "world")
  console.log(ngx.shared.myarray.items()[0])
}
```
Console output
```
2024/07/30 12:14:42 [info] 22#22: *1 js: ['hello','world']
```

### DIFF
--- a/ts/ngx_core.d.ts
+++ b/ts/ngx_core.d.ts
@@ -250,7 +250,7 @@ interface NgxFetchOptions {
 declare class SharedMemoryError extends Error {}
 
 type NgxSharedDictValue = string | number;
-type NgxKeyValuePair<V> = { key: string, value: V };
+type NgxKeyValuePair<V> = [string, V];
 
 /**
  * Interface of a dictionary shared among the working processes.


xeioex (changes requested):
We have our own commit log rules. The best way to get it is to `grep` commit history.

Proposed commit log:
```
Types: fixed NgxKeyValuePair definition.

NgxSharedDict.items() returns an array of `[key, value]` pairs, as `Object.entries()`.
```

ts/ngx_core.d.ts
```diff
@@ -250,7 +250,7 @@ interface NgxFetchOptions {
 declare class SharedMemoryError extends Error {}
 
 type NgxSharedDictValue = string | number;
-type NgxKeyValuePair<V> = { key: string, value: V };
+type NgxKeyValuePair<V> = [key: string, value: V];
```
xeioex, line 253: Also, it seems here we can remove names `key` and `value`.

To `type NgxKeyValuePair<V> = [string, V];`

TPXP:
Thanks for your feedback, I've reworded the commit and updated the type definition since adding names to array entries doesn't change much indeed (I hoped it would give some hints in IDEs but that does not seem to be the case).

xeioex (approved):

PR #769, opened 2024-07-30
feat(njs): expose capture group variables

TPXP:
### Proposed changes

This PR exposes capture group variables like `$1`, `$2` to njs scripts in the `Request.variables` object, avoiding duplicate parsing to determine this value (on nginx side and on njs side).

### Checklist

- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works
- [ ] If applicable, I have checked that any relevant tests pass after adding my changes

I'm not very familiar with the testing framework of this project so I couldn't test everything, here's my test scenario though (it makes sure we handle limits correctly and ensures previous variables are still accessible).
<details>
  <summary>Testing scenario</summary>
    
```javascript
// njs/test.js
function handler(r) {
  for (let i = 0; i < 10; i++) {
    console.log(`${i}: ${r.variables[i.toString()]}`);
  }
  console.log(`host: ${r.variables["http_host"]}`);
  r.return(200, `OK ${r.variables.http_host}`);
}

export default {handler}
```

```
# nginx.conf
daemon off;
worker_processes 1;

# Make sure JS log messages will appear on the console
error_log stderr info;

events {
  worker_connections 1024;
}

pid /tmp/nginx/run.pid;


http {
  access_log off;

  # Get nginx to start
  client_body_temp_path /tmp/nginx/body_tmp;
  proxy_temp_path /tmp/nginx/proxy_tmp;
  fastcgi_temp_path /tmp/nginx/fcgi_tmp;
  uwsgi_temp_path /tmp/nginx/uwsgi;
  scgi_temp_path /tmp/nginx/scgi;

  js_path "./njs/";
  js_import test from test.js;
  default_type  text/plain;

  server {
    listen       9090;
    # Define various locations with more or less matching groups
    location = /no_match {
      js_content test.handler;
    }
    location ~ ^/(.+)/(.+)/(.+)/(.+) {
      js_content test.handler;
    }
    location ~ ^/(.+)/(.+)/(.+) {
      js_content test.handler;
    }
    location ~ ^/(.+)/(.+) {
      js_content test.handler;
    }
    location ~ ^/(.+) {
      js_content test.handler;
    }
  }
}
```

Get URLS:
* /no_match
* /abc/def/ghi/jkl
* /abc/def/ghi
* /abc/def
* /abc

Logs look like this
```
2024/07/30 18:00:34 [info] 53869#0: *1 js: 0: /abc/def/ghi/jkl
2024/07/30 18:00:34 [info] 53869#0: *1 js: 1: abc
2024/07/30 18:00:34 [info] 53869#0: *1 js: 2: def
2024/07/30 18:00:34 [info] 53869#0: *1 js: 3: ghi
2024/07/30 18:00:34 [info] 53869#0: *1 js: 4: jkl
2024/07/30 18:00:34 [info] 53869#0: *1 js: 5: undefined
2024/07/30 18:00:34 [info] 53869#0: *1 js: 6: undefined
2024/07/30 18:00:34 [info] 53869#0: *1 js: 7: undefined
2024/07/30 18:00:34 [info] 53869#0: *1 js: 8: undefined
2024/07/30 18:00:34 [info] 53869#0: *1 js: 9: undefined
2024/07/30 18:00:34 [info] 53869#0: *1 js: host: localhost:9090
```
</details>

---
Inspired by https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_variable.c#L42

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -2954,10 +2954,10 @@ static njs_int_t
 ngx_http_js_request_variables(njs_vm_t *vm, njs_object_prop_t *prop,
     ngx_http_request_t *r, njs_value_t *setval, njs_value_t *retval)
 {
-    njs_int_t                   rc;
+    njs_int_t                   rc, is_capture, start, length;
     njs_str_t                   val, s;
     ngx_str_t                   name;
-    ngx_uint_t                  key;
+    ngx_uint_t                  i, key;
     ngx_http_variable_t        *v;
     ngx_http_core_main_conf_t  *cmcf;
     ngx_http_variable_value_t  *vv;
@@ -2972,6 +2972,31 @@ ngx_http_js_request_variables(njs_vm_t *vm, njs_object_prop_t *prop,
     name.len = val.length;
 
     if (setval == NULL) {
+        is_capture = 1;
+        for (i = 0; i < name.len; i++) {
+            if (name.data[i] < '0' || name.data[i] > '9') {
+                is_capture = 0;
+                break;
+            }
+        }
+
+        if (is_capture) {
+            key = ngx_atoi(name.data, name.len) * 2;
+            if (r->captures == NULL || r->captures_data == NULL
+                || r->ncaptures <= key)
+            {
+                njs_value_undefined_set(retval);
+                return NJS_DECLINED;
+            }
+
+            start = r->captures[key];
+            length = r->captures[key + 1] - start;
+
+            return ngx_js_prop(vm, njs_vm_prop_magic32(prop), retval,
+                               &r->captures_data[start], length);
+        }
+
+        /* Lookup the variable in nginx variables */
         key = ngx_hash_strlow(name.data, name.data, name.len);
 
         vv = ngx_http_get_variable(r, &name, key);
--- a/nginx/t/js_capture_variables.t
+++ b/nginx/t/js_capture_variables.t
@@ -1,0 +1,101 @@
+#!/usr/bin/perl
+
+# (C) Thomas P.
+
+# Tests for http njs module, reading location capture variables.
+
+###############################################################################
+
+use warnings;
+use strict;
+
+use Test::More;
+
+BEGIN { use FindBin; chdir($FindBin::Bin); }
+
+use lib 'lib';
+use Test::Nginx;
+
+###############################################################################
+
+select STDERR; $| = 1;
+select STDOUT; $| = 1;
+
+my $t = Test::Nginx->new()->has(qw/http/)
+	->write_file_expand('nginx.conf', <<'EOF');
+
+%%TEST_GLOBALS%%
+
+daemon off;
+
+events {
+}
+
+http {
+    %%TEST_GLOBALS_HTTP%%
+
+    js_import test.js;
+
+    server {
+        listen       127.0.0.1:8080;
+        server_name  localhost;
+
+        location /njs {
+            js_content test.njs;
+        }
+
+        location ~ /(.+)/(.+) {
+            js_content test.variables;
+        }
+    }
+}
+
+EOF
+
+$t->write_file('test.js', <<EOF);
+    function variables(r) {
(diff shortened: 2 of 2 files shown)


nginx/ngx_http_js_module.c
```diff
@@ -2972,6 +2972,38 @@ ngx_http_js_request_variables(njs_vm_t *vm, njs_object_prop_t *prop,
     name.len = val.length;
 
     if (setval == NULL) {
+        #if (NGX_PCRE)
+        is_capture = 1;
```
xeioex, line 2976: Nitpicking: is_capture = 0; as a starting point seems more appropriate.

nginx/ngx_http_js_module.c
```diff
@@ -2972,6 +2972,38 @@ ngx_http_js_request_variables(njs_vm_t *vm, njs_object_prop_t *prop,
...
+        for (size_t i = 0; i < name.len; i++) {
+            if (name.data[i] < '0' || name.data[i] > '9') {
+                is_capture = 0;
+                break;
+            }
+        }
+
+        if (is_capture) {
+            /*
```
xeioex, line 2985: The commit seems excessive here, because we have similar places with no such comments. It is a basic PCRE API knowledge.

xeioex (changes requested):
We also need autotests here. `nginx/t/js_capture_variables.t` can be added.
It can be based on `nginx/t/js_variables.t`. See `nginx/t/README` for more details.

Version check should be added as well:

```
TODO: {                                                                          
local $TODO = 'not yet' unless has_version('0.8.6');

<capture tests here>

}
```

xeioex (changes requested):
Proposed commit log:

`HTTP: expose capture group variables.`

TPXP (commented):
I've added a test, let me know if you have other remarks

nginx/ngx_http_js_module.c
xeioex: >The current logic is shorter, I think we should keep it like that.

Agree, disregard the comment.

xeioex:
Hi @TPXP,

I sent my minor improvements to your branch. Review them and squash the patches into a single one.
Otherwise looks good to me.

TPXP:
Thanks for the review, passing the index from the test simplifies the syntax - good idea. 👍 
I've just shortened the index fetching logic (to avoid converting the value to a number before converting it back to a string). https://github.com/nginx/njs/pull/769/files#diff-a5b346a0c232ffaa460fb68331af6fa52d12138b101fbcf5ec02bbea0fe0bca8R57

I've rebased and squashed commits, let me know if something else is missing.

nginx/ngx_http_js_module.c
```diff
@@ -2972,6 +2972,33 @@ ngx_http_js_request_variables(njs_vm_t *vm, njs_object_prop_t *prop,
     name.len = val.length;
 
     if (setval == NULL) {
+        #if (NGX_PCRE)
```
xeioex, line 2975: I just realized we do not need this #ifdef because njs module will not compile without nginx PCRE anyway.

TPXP:
By the way, I haven't received feedback yet for a companion nginx patch preventing duplicate calls to the js handler. Nothing urgent, but just to be sure could you confirm I've sent my patch at the right place to get it merged into nginx's core ? :)

https://mailman.nginx.org/pipermail/nginx-devel/2024-August/N7VFIYUKSZFUIAO24OJODKQGTP63R5NV.html

xeioex (changes requested):

nginx/ngx_http_js_module.c
TPXP: Ah, I haven't checked that. Removed the check then 👍

xeioex (approved):

xeioex:
@TPXP 

> I've sent my patch at the right place to get it merged into nginx's core ? :)

Yes the place is correct. The team received the patch and will give a reply as time permits. Unfortunately it may take a while because they are busy with more priority work.

xeioex:
Thank you for the patch, committed.

PR #771, opened 2024-08-07
feat(js_set): add nocache flag for uncacheable variables

TPXP:
### Proposed changes

This commit adds support for an additional `nocache` flag in `js_set` directives. If set, the resulting nginx variable will have no_cacheable set to 1. This enables us to dynamically recompute a variable if the context changed (for example, in case of an internal redirection). It also gives some control over the phase when variables are computed without needing to explain phases in the docs: as the variable is not cached, it is recomputed every time it is needed. Advanced users needing to hack something to get a JS code to run at a certain point in the config can simply refer to a nocached js_set variable in the corresponding directives and it is guaranteed the variable getter will be called when this directive evaluates the variable.

Should they want to benefit from this without performing too many calls, users can cache the result in a rewrite variable: `set $cached_variable $js_variable;`

I'm not very familiar with the testing framework of this project, but here's a test case.

<details>
  <summary>Test case</summary>
 
```
# nginx.conf - server block with njs loaded
  default_type  text/plain;

  js_path "./njs/";
  js_import test from test.js;
  js_set $test test.getTest nocache;

  server {
    listen       9090;

    location / {
      set $cached_test $test;
      if ($cached_test = "hello") {
        js_content test.redirect;
      }
      if ($cached_test != "hello") {
        return 200 $cached_test;
      }
    }
  }
```

```javascript
// njs/test.js
function getTest(r) {
  console.log('getTest call - URI is ' + r.uri);
  return "hello" + r.uri.substr(1);
}

function redirect(r) {
  r.internalRedirect('/test');
}

export default {getTest, redirect};
 ```

Expected result
```
curl 127.0.0.1:9090
hellotest
```

Expected logs
```
2024/08/07 12:06:52 [info] 40470#0: *1 js: getTest call - URI is /
2024/08/07 12:06:52 [info] 40470#0: *1 js: getTest call - URI is /test
```
JS calls may be duplicated due to what appears to be a bug in the way the rewrite handler evaluates complex expressions - I have sent [a patch](https://mailman.nginx.org/pipermail/nginx-devel/2024-August/N7VFIYUKSZFUIAO24OJODKQGTP63R5NV.html) to upstream nginx to address it.
</details>

### Checklist

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -342,7 +342,7 @@ static ngx_command_t  ngx_http_js_commands[] = {
       NULL },
 
     { ngx_string("js_set"),
-      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE2,
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE23,
       ngx_http_js_set,
       0,
       0,
@@ -1245,12 +1245,15 @@ static ngx_int_t
 ngx_http_js_variable_set(ngx_http_request_t *r, ngx_http_variable_value_t *v,
     uintptr_t data)
 {
-    ngx_str_t *fname = (ngx_str_t *) data;
+    ngx_js_set_t *vdata = (ngx_js_set_t *) data;
 
     ngx_int_t           rc;
     njs_int_t           pending;
+    ngx_str_t          *fname;
     njs_str_t           value;
     ngx_http_js_ctx_t  *ctx;
+
+    fname = &vdata->fname;
 
     rc = ngx_http_js_init_vm(r, ngx_http_js_request_proto_id);
 
@@ -1290,7 +1293,7 @@ ngx_http_js_variable_set(ngx_http_request_t *r, ngx_http_variable_value_t *v,
 
     v->len = value.length;
     v->valid = 1;
-    v->no_cacheable = 0;
+    v->no_cacheable = vdata->flags & NGX_NJS_VAR_NOCACHE;
     v->not_found = 0;
     v->data = value.start;
 
@@ -4639,7 +4642,8 @@ invalid:
 static char *
 ngx_http_js_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 {
-    ngx_str_t            *value, *fname, *prev;
+    ngx_str_t            *value;
+    ngx_js_set_t         *data, *prev;
     ngx_http_variable_t  *v;
 
     value = cf->args->elts;
@@ -4658,18 +4662,19 @@ ngx_http_js_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         return NGX_CONF_ERROR;
     }
 
-    fname = ngx_palloc(cf->pool, sizeof(ngx_str_t));
-    if (fname == NULL) {
+    data = ngx_palloc(cf->pool, sizeof(ngx_js_set_t));
+    if (data == NULL) {
         return NGX_CONF_ERROR;
     }
 
-    *fname = value[2];
+    data->fname = value[2];
+    data->flags = 0;
 
     if (v->get_handler == ngx_http_js_variable_set) {
-        prev = (ngx_str_t *) v->data;
+        prev = (ngx_js_set_t *) v->data;
 
-        if (fname->len != prev->len
-            || ngx_strncmp(fname->data, prev->data, fname->len) != 0)
+        if (data->fname.len != prev->fname.len
+            || ngx_strncmp(data->fname.data, prev->fname.data, data->fname.len) != 0)
         {
             ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
                                "variable \"%V\" is redeclared with "
@@ -4678,8 +4683,19 @@ ngx_http_js_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         }
     }
 
+    if (cf->args->nelts == 4) {
+        if (ngx_strcmp(value[3].data, "nocache") == 0) {
+            data->flags |= NGX_NJS_VAR_NOCACHE;
+
+        } else {
+            ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
+                "unrecognized flag \"%V\"", &value[3]);
+            return NGX_CONF_ERROR;
+        }
+    }
+
     v->get_handler = ngx_http_js_variable_set;
-    v->data = (uintptr_t) fname;
+    v->data = (uintptr_t) data;
 
     return NGX_CONF_OK;
 }
--- a/nginx/t/js_variables_nocache.t
+++ b/nginx/t/js_variables_nocache.t
@@ -1,0 +1,99 @@
+#!/usr/bin/perl
+
+# (C) Thomas P.
+
+# Tests for http njs module, setting non-cacheable nginx variables.
+
+###############################################################################
+
+use warnings;
+use strict;
+
+use Test::More;
+
+BEGIN { use FindBin; chdir($FindBin::Bin); }
+
+use lib 'lib';
+use Test::Nginx;
+
+###############################################################################
+
+select STDERR; $| = 1;
+select STDOUT; $| = 1;
+
+my $t = Test::Nginx->new()->has(qw/http rewrite/)
+	->write_file_expand('nginx.conf', <<'EOF');
+
+%%TEST_GLOBALS%%
+
+daemon off;
+
+events {
+}
+
+http {
+    %%TEST_GLOBALS_HTTP%%
+
+    js_set $nocache_var   test.variable nocache;
+    js_set $default_var   test.variable;
+    js_set $callcount_var test.callCount nocache;
+
+    js_import test.js;
+
+    server {
+        listen       127.0.0.1:8080;
+        server_name  localhost;
+
+        location /default_var {
+            set $a $default_var;
+            set $b $default_var;
+            return 200 '"$a/$b"';
+        }
+
+        location /nocache_var {
+            set $a $nocache_var;
+            set $b $nocache_var;
+            return 200 '"$a/$b"';
+        }
+
+        location /callcount_var {
+            set $a $callcount_var;
+            set $b $callcount_var;
+            return 200 '"$a/$b"';
+        }
+    }
+}
+
+EOF
+
+$t->write_file('test.js', <<EOF);
+    function variable(r) {
+        return Math.random().toFixed(16);
+    }
+
+    let n = 0;
+    function callCount(r) {
+        return (n += 1).toString();
+    }
+
+    export default {variable, callCount};
+
+EOF
+
+$t->try_run('no nocache njs variables')->plan(3);
+
+###############################################################################
+
+# We use backreferences to make sure the same value was returned for the two uses
+like(http_get('/default_var'), qr/"(.+)\/\1"/, 'cached variable');
+# Negative lookaheads don't capture, hence the .+ after it
+like(http_get('/nocache_var'), qr/"(.+)\/(?!\1).+"/, 'noncacheable variable');
+
+TODO: {
+    local $TODO = "Needs upstream nginx patch https://mailman.nginx.org/pipermail/nginx-devel/2024-August/N7VFIYUKSZFUIAO24OJODKQGTP63R5NV.html";
+
+    # Without the patch, this will give 2/4 (calls are duplicated)
+    like(http_get('/callcount_var'), qr/"1\/2"/, 'callcount variable');
+}
+
+###############################################################################
--- a/nginx/t/stream_js_variables_nocache.t
+++ b/nginx/t/stream_js_variables_nocache.t
@@ -1,0 +1,94 @@
+#!/usr/bin/perl
+
+# (C) Thomas P.
+
+# Tests for stream njs module, setting non-cacheable nginx variables.
+
+###############################################################################
+
+use warnings;
+use strict;
+
+use Test::More;
+
+BEGIN { use FindBin; chdir($FindBin::Bin); }
+
+use lib 'lib';
+use Test::Nginx;
+use Test::Nginx::Stream qw/ stream /;
+
+###############################################################################
+
+select STDERR; $| = 1;
+select STDOUT; $| = 1;
+
+my $t = Test::Nginx->new()->has(qw/stream stream_return/)
+	->write_file_expand('nginx.conf', <<'EOF');
+
+%%TEST_GLOBALS%%
+
+daemon off;
+
+events {
+}
+
+stream {
+    %%TEST_GLOBALS_STREAM%%
+
+    js_set $nocache_var   test.variable nocache;
+    js_set $default_var   test.variable;
+    js_set $callcount_var test.callCount nocache;
+
+    js_import test.js;
+
+    server {
+        listen  127.0.0.1:8081;
+        set     $a $default_var;
+        set     $b $default_var;
+        return  '"$a/$b"';
+    }
+
+    server {
+        listen  127.0.0.1:8082;
+        set     $a $nocache_var;
+        set     $b $nocache_var;
+        return  '"$a/$b"';
+    }
+
+    server {
+        listen  127.0.0.1:8083;
+        set     $a $callcount_var;
+        set     $b $callcount_var;
+        return  '"$a/$b"';
+    }
+}
+
+EOF
+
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: nginx/ngx_http_js_module.c [hunk 1]
```diff
@@ -76,6 +77,11 @@ typedef struct {
     uintptr_t              handler;
 }  ngx_http_js_header_t;
 
+typedef struct {
+    ngx_str_t fname;
+    unsigned flags;
+}  ngx_http_js_set_t;
```
xeioex, line 83: We have two modules: HTTP and stream. For symmetry this feature should be added to both modules.

Therefore put constant and `ngx_http_js_set_t` named as `ngx_js_set_t` to `nginx/ngx_js.h` just under `ngx_js_named_path_t` typefef.

nginx/ngx_http_js_module.c
```diff
@@ -1245,7 +1251,8 @@ static ngx_int_t
 ngx_http_js_variable_set(ngx_http_request_t *r, ngx_http_variable_value_t *v,
     uintptr_t data)
 {
-    ngx_str_t *fname = (ngx_str_t *) data;
+    ngx_http_js_set_t *vdata = (ngx_http_js_set_t *) data;
```
xeioex, line 1254: Style: sort lines in type length order, ie

```
ngx_str_t *fname = &vdata->fname;
ngx_js_set_t *vdata = (ngx_js_set_t *) data;
```

nginx/ngx_http_js_module.c
```diff
@@ -4653,8 +4662,18 @@ ngx_http_js_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         }
     }
 
+    if (cf->args->nelts == 4) {
+        if (ngx_strncmp(value[3].data, "nocache", value[3].len) == 0) {
+            data->flags |= NJS_VAR_NOCACHE;
```
xeioex, line 4667: Style: put empty line after `data->flags |= NJS_VAR_NOCACHE;`

nginx/ngx_http_js_module.c
```diff
@@ -4614,8 +4621,9 @@ ngx_http_js_periodic(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 static char *
 ngx_http_js_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 {
-    ngx_str_t            *value, *fname, *prev;
+    ngx_str_t            *value;
     ngx_http_variable_t  *v;
+    ngx_http_js_set_t    *data, *prev;
```
xeioex, line 4626: Style: sort var declaration in type length order.

nginx/ngx_http_js_module.c
```diff
@@ -47,6 +47,7 @@ typedef struct {
 #define NJS_HEADER_SINGLE      0x2
 #define NJS_HEADER_ARRAY       0x4
 
+#define NJS_VAR_NOCACHE 0x1
```
xeioex, line 50: Style: align constant value to other values in a file.

xeioex (changes requested):
Proposed commit log

```
Modules: added nocache flag for js_set variables.

This commit adds support for an additional `nocache` flag in `js_set`
directives.  If set, the resulting nginx variable will have no_cacheable flag set 
to 1. This enables to dynamically recompute a variable if the context changed
(for example, in case of an internal redirection).

In case of multiple calls in a location, users should cache the result in a
rewrite variable: `set $cached_variable $js_variable;`
```

nginx/ngx_http_js_module.c
TPXP: `fname` needs `vdata` defined (and casted). I've moved it to variables declarations and set it below instead. Hopefully it looks better now

nginx/t/js_variables_nocache.t
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/perl
+
+# (C) Dmitry Volyntsev
```
xeioex, line 3: Put the correct author here.

nginx/t/stream_js_variables_nocache.t
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/perl
+
+# (C) Dmitry Volyntsev
```
xeioex, line 3: Put the correct author here.

xeioex (changes requested):

TPXP (commented):
Thanks for the review, I've updated the author and description of tests

xeioex:
Hi @TPXP,

I see that tests fail randomly every 1 out of 8 runs when compiled with address sanitizer.

`--with-cc-opt="-DNGX_DEBUG_MALLOC=1 -DNGX_DEBUG_PALLOC=1 -DNJS_DEBUG_MEMORY=1 -fdiagnostics-color=auto -fsanitize=address -fno-sanitize=function"  --with-ld-opt=" -fsanitize=address"`

```
2024/08/15 10:55:58 [debug] 3571940#3571940: *2 http js variable call "test.variable"                                                2024/08/15 10:55:58 [error] 3571940#3571940: *2 js: variable call: 0.928769398946315length: 17             
2024/08/15 10:55:58 [debug] 3571940#3571940: *2 malloc: 000050300003CE50:17                            
2024/08/15 10:55:58 [debug] 3571940#3571940: *2 http js variable call "test.variable"                                               
2024/08/15 10:55:58 [error] 3571940#3571940: *2 js: variable call: 0.7905797879211605length: 18     
=================================================================       
==3571940==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50300003ce61 at pc 0xaaaad34d3d9c bp 0xffffd455e800 sp 0xffffd455dff0                                                                                                                        
WRITE of size 18 at 0x50300003ce61 thread T0                                                                         
    #0 0xaaaad34d3d98 in __asan_memcpy (/media/psf/workspace/nginx/nginx/objs/nginx+0x1b3d98) (BuildId: d5bbe7a8b19d077b7e82793f574356c6607f602e)                                                                                                                        
    #1 0xaaaad3747a34 in ngx_http_script_copy_var_code /media/psf/workspace/nginx/nginx/src/http/ngx_http_script.c:980:22        
    #2 0xaaaad38d191c in ngx_http_rewrite_handler /media/psf/workspace/nginx/nginx/src/http/modules/ngx_http_rewrite_module.c:180:9
    #3 0xaaaad36a3808 in ngx_http_core_rewrite_phase /media/psf/workspace/nginx/nginx/src/http/ngx_http_core_module.c:929:10        
    #4 0xaaaad36a3290 in ngx_http_core_run_phases /media/psf/workspace/nginx/nginx/src/http/ngx_http_core_module.c:875:14        
    #5 0xaaaad36a3020 in ngx_http_handler /media/psf/workspace/nginx/nginx/src/http/ngx_http_core_module.c:858:5              
    #6 0xaaaad36e823c in ngx_http_process_request /media/psf/workspace/nginx/nginx/src/http/ngx_http_request.c:2140:5
    #7 0xaaaad36f1168 in ngx_http_process_request_headers /media/psf/workspace/nginx/nginx/src/http/ngx_http_request.c:1529:13  
    #8 0xaaaad36ee4c4 in ngx_http_process_request_line /media/psf/workspace/nginx/nginx/src/http/ngx_http_request.c:1196:13        
    #9 0xaaaad36d9b04 in ngx_http_wait_request_handler /media/psf/workspace/nginx/nginx/src/http/ngx_http_request.c:523:5       
    #10 0xaaaad363c044 in ngx_epoll_process_events /media/psf/workspace/nginx/nginx/src/event/modules/ngx_epoll_module.c:901:17
    #11 0xaaaad35ee464 in ngx_process_events_and_timers /media/psf/workspace/nginx/nginx/src/event/ngx_event.c:248:12               
    #12 0xaaaad3631a54 in ngx_worker_process_cycle /media/psf/workspace/nginx/nginx/src/os/unix/ngx_process_cycle.c:721:9 
    #13 0xaaaad362773c in ngx_spawn_process /media/psf/workspace/nginx/nginx/src/os/unix/ngx_process.c:199:9                  
    #14 0xaaaad362db40 in ngx_start_worker_processes /media/psf/workspace/nginx/nginx/src/os/unix/ngx_process_cycle.c:344:9
    #15 0xaaaad362cbb0 in ngx_master_process_cycle /media/psf/workspace/nginx/nginx/src/os/unix/ngx_process_cycle.c:130:5       
    #16 0xaaaad35113ec in main /media/psf/workspace/nginx/nginx/src/core/nginx.c:384:9                                        
    #17 0xffffa62073f8 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #18 0xffffa62074c8 in __libc_start_main csu/../csu/libc-start.c:392:3
    #19 0xaaaad34358ec in _start (/media/psf/workspace/nginx/nginx/objs/nginx+0x1158ec) (BuildId: d5bbe7a8b19d077b7e82793f574356c6607f602e)

0x50300003ce61 is located 0 bytes after 17-byte region [0x50300003ce50,0x50300003ce61)
allocated by thread T0 here:
    #0 0xaaaad34d5d5c in malloc (/media/psf/workspace/nginx/nginx/objs/nginx+0x1b5d5c) (BuildId: d5bbe7a8b19d077b7e82793f574356c6607f602e)
    #1 0xaaaad3611574 in ngx_alloc /media/psf/workspace/nginx/nginx/src/os/unix/ngx_alloc.c:22:9
    #2 0xaaaad351fa28 in ngx_palloc_large /media/psf/workspace/nginx/nginx/src/core/ngx_palloc.c:220:9
    #3 0xaaaad351fd48 in ngx_pnalloc /media/psf/workspace/nginx/nginx/src/core/ngx_palloc.c:144:12
    #4 0xaaaad3751448 in ngx_http_script_complex_value_code /media/psf/workspace/nginx/nginx/src/http/ngx_http_script.c:1778:19
    #5 0xaaaad38d191c in ngx_http_rewrite_handler /media/psf/workspace/nginx/nginx/src/http/modules/ngx_http_rewrite_module.c:180:9  
    #6 0xaaaad36a3808 in ngx_http_core_rewrite_phase /media/psf/workspace/nginx/nginx/src/http/ngx_http_core_module.c:929:10
    #7 0xaaaad36a3290 in ngx_http_core_run_phases /media/psf/workspace/nginx/nginx/src/http/ngx_http_core_module.c:875:14
    #8 0xaaaad36a3020 in ngx_http_handler /media/psf/workspace/nginx/nginx/src/http/ngx_http_core_module.c:858:5
    #9 0xaaaad36e823c in ngx_http_process_request /media/psf/workspace/nginx/nginx/src/http/ngx_http_request.c:2140:5
    #10 0xaaaad36f1168 in ngx_http_process_request_headers /media/psf/workspace/nginx/nginx/src/http/ngx_http_request.c:1529:13
    #11 0xaaaad36ee4c4 in ngx_http_process_request_line /media/psf/workspace/nginx/nginx/src/http/ngx_http_request.c:1196:13
    #12 0xaaaad36d9b04 in ngx_http_wait_request_handler /media/psf/workspace/nginx/nginx/src/http/ngx_http_request.c:523:5
    #13 0xaaaad363c044 in ngx_epoll_process_events /media/psf/workspace/nginx/nginx/src/event/modules/ngx_epoll_module.c:901:17
    #14 0xaaaad35ee464 in ngx_process_events_and_timers /media/psf/workspace/nginx/nginx/src/event/ngx_event.c:248:12
    #15 0xaaaad3631a54 in ngx_worker_process_cycle /media/psf/workspace/nginx/nginx/src/os/unix/ngx_process_cycle.c:721:9
    #16 0xaaaad362773c in ngx_spawn_process /media/psf/workspace/nginx/nginx/src/os/unix/ngx_process.c:199:9
    #17 0xaaaad362db40 in ngx_start_worker_processes /media/psf/workspace/nginx/nginx/src/os/unix/ngx_process_cycle.c:344:9
    #18 0xaaaad362cbb0 in ngx_master_process_cycle /media/psf/workspace/nginx/nginx/src/os/unix/ngx_process_cycle.c:130:5
    #19 0xaaaad35113ec in main /media/psf/workspace/nginx/nginx/src/core/nginx.c:384:9
    #20 0xffffa62073f8 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #21 0xffffa62074c8 in __libc_start_main csu/../csu/libc-start.c:392:3
    #22 0xaaaad34358ec in _start (/media/psf/workspace/nginx/nginx/objs/nginx+0x1158ec) (BuildId: d5bbe7a8b19d077b7e82793f574356c6607f602e)

SUMMARY: AddressSanitizer: heap-buffer-overflow (/media/psf/workspace/nginx/nginx/objs/nginx+0x1b3d98) (BuildId: d5bbe7a8b19d077b7e82793f574356c6607f602e) in __asan_memcpy
Shadow bytes around the buggy address:
  0x50300003cb80: 00 00 00 00 fa fa fd fd fd fd fa fa 00 00 00 00
  0x50300003cc00: fa fa 00 00 00 00 fa fa 00 00 00 00 fa fa 00 00
  0x50300003cc80: 00 00 fa fa 00 00 00 00 fa fa 00 00 00 fa fa fa
  0x50300003cd00: fd fd fd fa fa fa fd fd fd fa fa fa fd fd fd fa
  0x50300003cd80: fa fa fd fd fd fa fa fa fd fd fd fa fa fa 00 00
=>0x50300003ce00: 00 fa fa fa 00 00 00 fa fa fa 00 00[01]fa fa fa
  0x50300003ce80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x50300003cf00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x50300003cf80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x50300003d000: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x50300003d080: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa                                                                                                        
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==3571940==ABORTING
```



The root cause is that variable handler is called twice for the length and the value, when the value is larger than the allocated length we have a heap-buffer-overflow. That is related to [your](https://mailman.nginx.org/pipermail/nginx-devel/2024-August/N7VFIYUKSZFUIAO24OJODKQGTP63R5NV.html) patch. The patch seems to fix the issue and I think we need to solve the issue on the nginx side first before commiting this change to njs.

nginx/ngx_http_js_module.c
```diff
@@ -4678,8 +4683,19 @@ ngx_http_js_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         }
     }
 
+    if (cf->args->nelts == 4) {
+        if (ngx_strncmp(value[3].data, "nocache", value[3].len) == 0) {
```
xeioex, line 4687: check here the exact length: 7.

```
if (ngx_strcmp(value[3].data, "nocache") == 0) {
```

Otherwise partial matches will work too. The same with stream module.

xeioex (changes requested):

TPXP:
The buffer overflow indeed makes sense since a stringified `Math.random` value would return a string of variable length (trailing zeros would be omitted). For example, if successive calls of Math.random return `0.25738823928921` and `0.89781232223590`, ngx_strlen would compute the length of `"0.25738823928921"` (16) and try to copy 16 bytes from `"0.8978123222359"` (15 bytes without the trailing zero) so we have an out of bounds read.

Thanks for giving the additional context on the ngx-devel mailing list, let's wait for my patch to make its way through first then 👍

nginx/ngx_http_js_module.c
TPXP: Thanks, I was worried about out of bounds reading but other parts of nginx's source code show this is the right way 👍

xeioex:
Hi @TPXP,

I propose the adjust the tests for a variable to have fixed length, we can also document this limitation.
It is a serious limitation for your use case?

TPXP:
Sure, we can do this @xeioex . I've updated the tests to return a fixed-size random string with `toFixed()`

TPXP:
We may consider adding some sort of `TODO` test checking if the handler is not called twice (we can use a global JS variable in the handler for this)
```js
let callCount = 0;
function variable(r) {
    return (callCount+=1).toString();
}
```

xeioex (approved):

PR #772, opened 2024-08-09
Buffer module for QuickJS.

xeioex:


VadimZhestikov (approved):

VadimZhestikov (approved):

VadimZhestikov (approved):

src/njs_buffer.c
```diff
@@ -1618,6 +1618,8 @@ njs_buffer_write_string(njs_vm_t *vm, njs_value_t *value,
         goto done;
     }
 
+    length = njs_min(str.length, (size_t) length);
```
VadimZhestikov, line 1718: This change does not cover edge case when there is no room for trailing multibyte character in a buffer,
        for example:
            var buf = Buffer.alloc(10);
            var len = buf.write('\u00bd', 0, 1);
        It should set len == 0, and write nothing to the buf.

        For referenece see quotation: "However, partially encoded characters will not be written."
        from https://nodejs.org/api/buffer.html#bufwritestring-offset-length-encoding

VadimZhestikov (changes requested):
Looks good, except for small comment.

VadimZhestikov (approved):

src/qjs_buffer.c
```diff
@@ -103,14 +158,104 @@ static const JSCFunctionListEntry qjs_buffer_export[] = {
...
+    JS_CFUNC_DEF("concat", 1, qjs_buffer_concat),
     JS_CFUNC_DEF("from", 3, qjs_buffer_from),
     JS_CFUNC_DEF("isBuffer", 1, qjs_buffer_is_buffer),
+    JS_CFUNC_DEF("isEncoding", 1, qjs_buffer_is_encoding),
 };
 
 
 static const JSCFunctionListEntry qjs_buffer_proto[] = {
+    JS_CFUNC_DEF("compare", 5, qjs_buffer_prototype_compare),
+    JS_CGETSET_DEF("constructor", qjs_buffer_prototype_constructor, NULL),
```
VadimZhestikov, line 174: constructor of Buffer proto  need to be usual "configurable, writable, not enumerable" property,
which is initially set to Buffer.

Example of issue:

  var b = Buffer.from('a');
  var buffer_old = Buffer;
  Buffer = 1;

as result got wrong::

  b.constructor === 1;

while should produce:

   b.constructor === buffer_old;;

VadimZhestikov (changes requested):
JS_CGETSET_DEF("constructor", qjs_buffer_prototype_constructor, NULL),

It need to be usual "configurable, writable, not enumerable" property

src/qjs_buffer.c
VadimZhestikov: Otherwise all parts of the PR looks good to me.

VadimZhestikov (approved):
Looks good.

PR #778, opened 2024-08-31
Preliminary patch set before QuickJS modules.

xeioex:


VadimZhestikov (commented):
See comment about removing Function() together with eval().
Otherwise looks good.

VadimZhestikov (approved):
Looks good

PR #800, opened 2024-10-15
Fixing flaky ASAN CI build.

xeioex:


VadimZhestikov (approved):
Both commits looks good. "Lazy stack symbolization" nice catch.

Issue #698, opened 2024-03-06
Introducing QuickJS engine as alternative JS engine.


xeioex:
[QuickJS](https://github.com/bellard/quickjs) will be added as an alternative JS engine. 
QuickJS is supported and developed by a wider community, it has almost full ES spec coverage.

QuickJS will be added in stages:

- [X] to njs shell
- [ ] as an alternative JS engine for nginx modules. 
- [ ] all built-in njs modules will be ported to QuickJS

NJS engine will stay a default JS engine for a while for backward compatibility (at least until QuickJS version is mature enough).

radsoc:
I am very concerned about this move.
It contradicts everything that was said in the 2018 talk (https://youtu.be/Jc_L6UffFOs), which I fully support.
Performance is key; full ES spec coverage is not; njs is not a nodejs replacement.

![vlcsnap-2024-03-08-10h23m13s249](https://github.com/nginx/njs/assets/6185175/e7054cb6-a0a4-4861-8bfd-3e41a7fb2b9f)
![vlcsnap-2024-03-08-10h29m04s985](https://github.com/nginx/njs/assets/6185175/a7fa5550-4c44-4c61-929f-2ea4b6484744)
![vlcsnap-2024-03-08-10h06m45s269](https://github.com/nginx/njs/assets/6185175/8a424c14-88d5-4f42-af10-f3e21c1a547c)
![vlcsnap-2024-03-08-10h52m30s465](https://github.com/nginx/njs/assets/6185175/e1823007-6a38-492b-bd68-4c2e387e6a84)
![vlcsnap-2024-03-08-10h33m30s382](https://github.com/nginx/njs/assets/6185175/709fe5f2-798c-4ec9-8cc4-2e58a5d3beb4)

QuickJS and Ducktape are in the same ballpark.

<img width="663" alt="2024-03-08_10-17" src="https://github.com/nginx/njs/assets/6185175/d757a04d-e545-40a7-b3e8-49a0c18d41af">

And QuickJS is garbage collected.

<img width="839" alt="2024-03-08_10-38" src="https://github.com/nginx/njs/assets/6185175/dafee600-2156-4045-a464-699d51b3e0cb">

xeioex:
Hi @radsoc,

Thank you for the feedback and for letting us know that NJS is preferred. Let me address some of your questions.

> Performance is key; full ES spec coverage is not; NJS is not a Node.js replacement.

[...]

PR #808, opened 2024-10-28
Implement `process.kill()` function

stefansundin:
### Proposed changes

This PR implements the `process.kill()` function, using the same API as Node.js. https://nodejs.org/api/process.html#processkillpid-signal

My use case for this function is not to kill processes, but to send other signals, like `SIGHUP`. But the default is to send `SIGTERM`, just like in Node.js.

I am not 100% sure that I am using the correct string functions to compare the input with the signal names. It is working but I feel like it is not the proper way to do it. Please review the patch and let me know what the proper way of doing things are.

I am using `#ifdef` for almost all signal names, except for `SIGTERM`. Not sure if it is strictly necessary but I don't think all signals are defined for all platforms?

I tested this on Debian 12.

### Checklist


src/njs_builtin.c
```diff
@@ -1380,6 +1381,186 @@ njs_process_object_ppid(njs_vm_t *vm, njs_object_prop_t *prop,
...
+        njs_value_string_get(arg, &str);
+
+        if (njs_strncmp(str.start, "SIGTERM", njs_length("SIGTERM")) == 0) {
+          signal = SIGTERM;
+        }
+        #ifdef SIGABRT
+        else if (njs_strncmp(str.start, "SIGABRT", njs_length("SIGABRT")) == 0) {
+          signal = SIGABRT;
+        }
+        #endif
```
stefansundin, line 1412: I'd love to hear if anyone knows a more compact way of doing this.

test/shell_test.exp
```diff
@@ -423,6 +423,8 @@ njs_run {"-c" "console.log(process.pid)"} "\\d+"
 
 njs_run {"-c" "console.log(process.ppid)"} "\\d+"
 
+njs_run {"-c" "console.log(process.kill(process.pid, 0))"} "true"
```
stefansundin, line 426: The tests runs `kill(pid, 0)` which just checks that the process exists. It doesn't kill anything.

ts/njs_core.d.ts
```diff
@@ -579,6 +579,11 @@ interface NjsProcess {
     readonly ppid: number;
     readonly argv: string[];
     readonly env: NjsEnv;
+
+    /**
+     * @since 0.8.8
```
stefansundin, line 584: Here I'm just optimistic that this will be merged before the next version. 😄

xeioex:
Hi @stefansundin,

Thank you for the patch. We need the [QJS](https://github.com/nginx/njs/issues/698) implementation as well for shell tests to pass. I will do that myself, after polishing you patch.

For signals, I propose to use only standard signals from `man 7 signal` up to P1990. With that we can get rid of ifdefs.
For `else if` blocks, may be the better way is to use a table. Something like [this](https://github.com/nginx/njs/blob/master/external/njs_fs_module.c#L245).

stefansundin:
Thank you @xeioex. I really appreciate the assistance.

I have pushed a new commit that I think incorporates your suggestions. Please let me know if there's anything you want me to change.

I can squash the commits if you prefer. Just let me know.

xeioex:
@stefansundin 

Thank you for the fast response. 

1) Looking at we we have now, I think we can remove "SIG" part from the table, so we do not check it again and again. Instead we can check for "SIG" just once, before the table.

2) Please, change `// P1990 signals from `man 7 signal` are supported`
to `/* ... */` as we do not use `//`-style comments. 

3) Also, feel free to squash all the changes into a single patch.

stefansundin:
I think I managed to do it correctly. It got a little bit tricky with the pointers since I don't think there is a helper function to do these comparisons. But it is working and there are length checks so I think it is secure.

test/shell_test.exp
stefansundin, line 426: Should I comment this line out?

I guess it didn't run with my local `make test`, probably because I didn't have QuickJS installed.

xeioex:
Hi @stefansundin,

I added QuickJS part. I also added `minor improvements` and `Shell: added process.kill() test` patches.
Please test them, and if you like them squash to your original commit.

stefansundin:
I reviewed it and it looks great. 👍 

I rebuilt my project and tested with the new changes and it worked.

I tried to build with QuickJS to test that but I encountered a compilation issue (building in docker):

```
37.66 /usr/bin/ld: /root/quickjs/libquickjs.a(quickjs.o): relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol `js_realloc_rt' which may bind externally can not be used when making a shared object; recompile with -fPIC
37.66 /root/quickjs/quickjs.c:1458:(.text+0x3fc90): dangerous relocation: unsupported relocation
```

In any case, I squashed the commits since I think you probably tested QuickJS. :)

xeioex:
@stefansundin 

>I tried to build with QuickJS to test that but I encountered a compilation issue (building in docker):

Something like `CFLAGS='-fPIC' make libquickjs.a` should fix the problem.

>In any case, I squashed the commits since I think you probably tested QuickJS. :)

Thank you for the contribution. Next week @VadimZhestikov  will do the review of QuickJS patches, and will merge the PR.

stefansundin:
Thanks for the `-fPIC` tip, I was able to built QuickJS with it. 👍 

However, I think think there may be an issue in the QuickJS implementation, I got this error when attempting to call `process.exit()`:

```
[error] 239#239: *1 js call exception: ReferenceError: 'process' is not defined
```

VadimZhestikov (approved):
Looks good

xeioex:
@stefansundin Thank you, committed

PR #820, opened 2024-11-15
Reimplemented js_preload_object without a separate VM.

xeioex:


nginx/ngx_js.c
```diff
@@ -3271,30 +3252,18 @@ ngx_js_preload_object(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
-    options.addons = njs_js_addon_modules_shared;
-
-    vm = njs_vm_create(&options);
-    if (vm == NULL) {
-        goto error;
-    }
-
     njs_str_t str = njs_str(
-        "import fs from 'fs';"
+        "import __fs from 'fs';"
```
VadimZhestikov, line 3266: Looks like no need rename fs, its scope is current module only.

src/njs_object.c
```diff
@@ -917,7 +917,7 @@ njs_get_own_ordered_keys(njs_vm_t *vm, const njs_object_t *object,
     njs_lvlhsh_each_init(&lhe, &njs_object_hash_proto);
     hash = &object->shared_hash;
 
-    for ( ;; ) {
+    for ( ; !(flags & NJS_ENUM_NON_SHARED_ONLY); ) {
```
VadimZhestikov, line 920: this expression is const in the loop, better use explicit if before loop:

if (!(flags & NJS_ENUM_NON_SHARED_ONLY))
for ( ; ; ) {

src/njs_builtin.c
```diff
@@ -920,6 +920,8 @@ njs_top_level_constructor(njs_vm_t *vm, njs_object_prop_t *self,
         ctor = &njs_vm_ctor(vm, njs_prop_magic16(self));
 
         njs_set_function(retval, ctor);
+
+        return NJS_OK;
```
VadimZhestikov, line 924: Looks like it introduce performance decrease for all read accesses to all top_level_constructors

VadimZhestikov (commented):
Looks good except for comments

nginx/ngx_js.c
xeioex: No, it is in the global scope.

if there is a js_import directive as follow

`js_import fs from lib.js` or `js_import fs.js`

we will get 
`SyntaxError: "fs" has already been declared `.

src/njs_object.c
xeioex: Will do a check with goto before. But I expect the effect of the change unmeasurable. Also wanted to reduce the diff, while preserving the style.

src/njs_object.c
```diff
@@ -1187,6 +1193,82 @@ njs_traverse_visited(njs_arr_t *list, const njs_value_t *value)
...
+
+    for ( ;; ) {
+
+        if (s->index >= s->keys->length) {
+            njs_flathsh_init(&njs_object(&s->value)->hash);
+            njs_object(&s->value)->shared = 1;
+            njs_array_destroy(vm, s->keys);
+            s->keys = NULL;
+
+            if (s == &state[0]) {
```
VadimZhestikov, line 1237: s is const and it is always == &state[0]

src/njs_object.c
```diff
@@ -1187,6 +1193,82 @@ njs_traverse_visited(njs_arr_t *list, const njs_value_t *value)
...
+njs_int_t
+njs_object_make_shared(njs_vm_t *vm, njs_object_t *object)
+{
+    njs_int_t             ret;
+    njs_arr_t             visited;
+    njs_object_t          **start;
+    njs_value_t           *key;
+    njs_traverse_t        *s;
+    njs_property_query_t  pq;
+    njs_traverse_t        state[NJS_TRAVERSE_MAX_DEPTH];
```
VadimZhestikov, line 1205: Looks like this is just single value, not array

src/njs_object.c
```diff
@@ -1187,6 +1193,82 @@ njs_traverse_visited(njs_arr_t *list, const njs_value_t *value)
...
+
+    s = &state[0];
+    s->parent = NULL;
```
VadimZhestikov, line 1208: s->parent set but not used?

src/njs_object.c
```diff
@@ -1187,6 +1193,82 @@ njs_traverse_visited(njs_arr_t *list, const njs_value_t *value)
...
+    pq.lhq.replace = 0;
+    pq.lhq.pool = vm->mem_pool;
+
+    for ( ;; ) {
+
+        if (s->index >= s->keys->length) {
+            njs_flathsh_init(&njs_object(&s->value)->hash);
+            njs_object(&s->value)->shared = 1;
+            njs_array_destroy(vm, s->keys);
+            s->keys = NULL;
```
VadimZhestikov, line 1235: s->key set, but never used then

src/njs_object.c
xeioex: I did a full and recursive version, which we do not need in our case, because we make our preload object constant by construction. But in the future, we can even allow to modify preload objects per cloned VM without changing the shared part.

With a recursive version `s` is not constant, and `state`, `s->parent` are used.

VadimZhestikov (approved):
Looks good

PR #822, opened 2024-11-26
HTTP: fixed buffered output.

xeioex:


mrajagopal:
I built NJS master + patch and tested it on nginx version: nginx/1.25.5 (nginx-plus-r32-p1).
I can confirm that the patch addresses the issue that initially identified this bug.

VadimZhestikov (approved):

PR #829, opened 2024-12-14
FS module for QuickJS engine.

xeioex:


external/qjs_fs_module.c
```diff
@@ -0,0 +1,3076 @@
...
+    if (JS_DefinePropertyValueStr(cx, obj, "name", name,
+                                  JS_PROP_ENUMERABLE) < 0)
+    {
+        JS_FreeValue(cx, obj);
+        return JS_EXCEPTION;
+    }
+
+    if (entry != NULL) {
+        type = JS_NewInt32(cx, qjs_dentry_type(entry));
+        if (JS_IsException(type)) {
```
VadimZhestikov, line 1095: JS_NewInt32 never produce exception, so no need test for it here and in other places.

VadimZhestikov (approved):
Looks good

PR #834, opened 2025-01-04
Modules: removed extra VM creation per server.

xeioex:


VadimZhestikov (approved):
Looks good, is it still draft?

PR #841, opened 2025-01-18
QuickJS: added WebCrypto module.

xeioex:


nginx/t/stream_js_webcrypto.t
```diff
@@ -0,0 +1,83 @@
...
+
+    function random_values_test(s) {
+        let buf = new Uint32Array(32);
+        crypto.getRandomValues(buf);
+        let bits1 = buf.reduce((a, v)=> a + count1(v), 0);
+        let nbits = buf.length * 32;
+        let mean = nbits / 2;
+        let stdd = Math.sqrt(nbits / 4);
+
+        return bits1 > (mean - 10 * stdd) && bits1 < (mean + 10 * stdd);
```
VadimZhestikov, line 71: While bits1 outside of (mean+-10σ) has _extremally_ low probability, such value is absolutely correct. May be, to avoid confusions, better replace this test by another one with exactly known result?

VadimZhestikov (dismissed):
Looks good

nginx/t/stream_js_webcrypto.t
VadimZhestikov: This is very minor note, can be kept as is.

VadimZhestikov (approved):
Looks good

PR #850, opened 2025-02-08
Modules: fix writing to read-only memory.

xeioex:


nginx/ngx_http_js_module.c
```diff
@@ -6184,9 +6202,18 @@ ngx_http_qjs_variables_own_property(JSContext *cx, JSPropertyDescriptor *pdesc,
         return 1;
     }
 
-    key = ngx_hash_strlow(name.data, name.data, name.len);
+    name_lc.data = (name.len < sizeof(storage))
```
hongzhidao, line 6205: It would be great if we support VLA(variable length array), I mean do it like this.
```
storage  [val.length];
```
Then we can get rid of memory allocation.

hongzhidao (dismissed):
The fix looks good.
It looks like doing something like `preserver {some var related stuff} case sensitivity`.
I mean we could make the commit title clearer.

nginx/ngx_http_js_module.c
xeioex: VLA are not very portable. 
Also, fixed buffer size helps us to ensure the more predictable memory consumption.

xeioex:
@hongzhidao 

1) added extended commit log
2) removed one comparison from the fast code pass
3) added forgotten stack storage for `ngx_stream_qjs_variables_set_property()`

hongzhidao (approved):
LGTM.

PR #867, opened 2025-03-18
QuickJS: fixed ngx_qjs_string().

hongzhidao:
Fix the function to properly handler strings like "\0" which happens in tests. In js_fetch_objects.t:
```
for (var i = 0; i < 128; i++) {
  var c = String.fromCodePoint(i);
```

Refactored out from qjs fetch patches.

### DIFF
--- a/nginx/ngx_js.c
+++ b/nginx/ngx_js.c
@@ -1482,12 +1482,10 @@ ngx_qjs_string(JSContext *cx, JSValueConst val, ngx_str_t *dst)
 
 string:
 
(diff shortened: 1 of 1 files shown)


xeioex (changes requested):
Suggested commit log:

```
QuickJS: fixed ngx_qjs_string() to handle strings containing "\0".
```

Otherwise Looks good.

xeioex (approved):

xeioex (approved):

PR #870, opened 2025-03-19
QuickJS: using JS_AddIntrinsicBigInt() if detected.

VadimZhestikov:
In quickjs JS_AddIntrinsicBigInt() is moved from public API to JS_AddIntrinsicBaseObjects():

commit 6d6893bfa3d383d4952b67954003800aba1f4be8
Author: Fabrice Bellard <fabrice@bellard.org>
Date:   Wed Mar 19 12:33:54 2025 +0100

    fixed BigInt hashing - removed -fno-bigint in qjsc and JS_AddIntrinsicBigInt() (BigInt is now considered as a base object)

### DIFF
--- a/auto/quickjs
+++ b/auto/quickjs
@@ -137,6 +137,29 @@ if [ $NJS_TRY_QUICKJS = YES ]; then
 
         . auto/feature
 
+        njs_feature="QuickJS JS_AddIntrinsicBigInt()"
+        njs_feature_name=NJS_HAVE_QUICKJS_ADD_INTRINSIC_BIG_INT
+        njs_feature_test="#if defined(__GNUC__) && (__GNUC__ >= 8)
+                          #pragma GCC diagnostic push
+                          #pragma GCC diagnostic ignored \"-Wcast-function-type\"
+                          #endif
+
+                          #include <quickjs.h>
+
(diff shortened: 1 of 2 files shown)


xeioex:
The current approach seems unreliable and depends on the version of the libquickjs.a or NG version.
Let's do the configure check for `JS_AddIntrinsicBigInt()`  [here](https://github.com/nginx/njs/blob/master/auto/quickjs#L117).

xeioex (commented):
The current approach seems unreliable and depends on the exact version of the libquickjs.a or NG version.
Let's do the configure check for JS_AddIntrinsicBigInt() [here](https://github.com/nginx/njs/blob/master/auto/quickjs#L117).

xeioex (changes requested):
The current approach seems unreliable and depends on the exact version of the libquickjs.a or NG version.
Let's do the configure check for JS_AddIntrinsicBigInt() [here](https://github.com/nginx/njs/blob/master/auto/quickjs#L117).

xeioex (dismissed):
Suggested commit log:

`QuickJS: using JS_AddIntrinsicBigInt() if detected. `

Otherwise looks good.

xeioex (approved):

PR #864, opened 2025-03-17
qjs fetch.

hongzhidao:

PR #873, opened 2025-03-24
QuickJS: added missed JS_NewClass() for the SharedDictError class.

hongzhidao:
See https://github.com/nginx/njs/pull/864#issuecomment-2746931068


xeioex (approved):
Looks good.

PR #886, opened 2025-04-10
Fetch: refactored out ngx_js_http_resolve().

hongzhidao:
Prepared for related ngx_js_http_* patch.


nginx/ngx_js_fetch.c
```diff
@@ -1330,8 +1319,42 @@ ngx_js_http_alloc(njs_vm_t *vm, ngx_pool_t *pool, ngx_log_t *log)
...
+    ctx = ngx_resolve_start(r, NULL);
+    if (ctx == NULL) {
+        return NULL;
+    }
+
+    if (ctx == NGX_NO_RESOLVER) {
+        return ctx;
+    }
+
+    http->ctx = ctx;
```
xeioex, line 1338: We can move it to the end of the function, so `http->ctx = NULL;` will not be needed.

xeioex (approved):
Otherwise looks good.

PR #888, opened 2025-04-10
Njs fetch cleanup

hongzhidao:
Prepared for related ngx_js_http_* patch.

### DIFF
--- a/nginx/ngx_js_fetch.c
+++ b/nginx/ngx_js_fetch.c
@@ -277,6 +277,16 @@ static njs_int_t ngx_fetch_flag_set(njs_vm_t *vm, const ngx_js_entry_t *entries,
 static njs_int_t ngx_js_fetch_init(njs_vm_t *vm);
 
 
+njs_inline void
+ngx_js_http_resolve_done(ngx_js_http_t *http)
+{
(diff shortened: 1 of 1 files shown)


nginx/ngx_js_fetch.c
```diff
@@ -277,6 +277,16 @@ static njs_int_t ngx_fetch_flag_set(njs_vm_t *vm, const ngx_js_entry_t *entries,
 static njs_int_t ngx_js_fetch_init(njs_vm_t *vm);
 
 
+njs_inline void
```
xeioex, line 281: Let's make it an ordinary static function and move somewhere down. 
It is up to compiler to inline it.

nginx/ngx_js_fetch.c
```diff
@@ -287,6 +288,16 @@ ngx_js_http_resolve_done(ngx_js_http_t *http)
 }
 
 
+njs_inline void
+ngx_js_http_close_peer(ngx_js_http_t *http)
```
xeioex, line 292: Let's make it an ordinary static function and move somewhere down. 
It is up to compiler to inline it.

xeioex (changes requested):
Otherwise looks good.

PR #894, opened 2025-04-21
Test262: fixed mktemp format on OpenBSD.

xeioex:
mktemp: insufficient number of Xs in template `/tmp/njs_test.XXX'


VadimZhestikov (approved):
Looks good

PR #871, opened 2025-03-20
Tag some char arrays with __attribute__((__nonstring__))

ac000:
GCC 15 implements a new warning `-Wunterminated-string-initialization` that is also enabled by `-Wextra` that we enable.

This causes compilation failures (due to -Werror) due to the likes of

```c
static const u_char  hex[16] = "0123456789ABCDEF";
```

E.g.

```
src/njs_string.h: In function ‘njs_string_encode’:
src/njs_string.h:212:36: error: initializer-string for array of ‘unsigned char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (17 chars into 16 available) [-Werror=unterminated-string-initialization]
  212 |     static const u_char  hex[16] = "0123456789ABCDEF";
      |                                    ^~~~~~~~~~~~~~~~~~
```

These are very much meant not to be NUL terminated.

Now we could just disable this new warning. But I think it is worth leaving it enabled (the GCC developers also obviously feel it's useful enough to enable under -Wexta), anything that helps the compiler help us avoid silly mistakes is a good thing(tm), particularly in the current climate.

So rather than disable this warning, we can make use of the GCC "nonstring" variable attribute `__attribute__((__nonstring__))`. 

This attribute is used to mark character arrays that are intentionally not NUL terminated.

So the above example would become (we of course wrap it in a more friendly name `NJS_NONSTRING`)

```c
static const u_char  hex[16] NJS_NONSTRING = "0123456789ABCDEF";
```

* The first commit adds a wrapper around `__attribute__((__nonstring__))`

* The second commit tags various character arrays with this attribute.

Note: This was done with an eye towards the atomic strings work which removes the need for one of the attribute tags.
 
```
The following changes since commit d34fcb03cf2378a644a3c7366d58cbddc2771cbd:

  HTTP: fixed r.subrequest() error handling. (2024-06-12 15:09:21 -0700)

are available in the Git repository at:

  git@github.com:ac000/njs.git wusi

for you to fetch changes up to a00cc6d3897393f723d9f11f78651f4cbcb8cc76:

  Tag various character arrays with NJS_NONSTRING (2025-03-20 17:32:58 +0000)

----------------------------------------------------------------
Andrew Clayton (2):
      auto/clang: Add a NJS_NONSTRING macro
      Tag various character arrays with NJS_NONSTRING

 auto/clang        | 14 ++++++++++++++
 src/njs_clang.h   |  8 ++++++++
 src/njs_sprintf.c |  4 ++--
[...]

PR #902, opened 2025-05-02
Fixed GCC 15 build with -Wunterminated-string-initialization.

xeioex:
In file included from src/njs_main.h:48,
                 from src/njs_diyfp.c:12:
src/njs_string.h: In function ‘njs_string_encode’: src/njs_string.h:229:36: error: initializer-string for array of ‘unsigned char’
	truncates NUL terminator but destination lacks ‘nonstring’ attribute (
	17 chars into 16 available) [-Werror=unterminated-string-initialization]
  229 |     static const u_char  hex[16] = "0123456789ABCDEF";

### DIFF
--- a/external/qjs_query_string_module.c
+++ b/external/qjs_query_string_module.c
@@ -537,7 +537,7 @@ qjs_string_encode(const uint32_t *escape, size_t size, const u_char *src,
     u_char *dst)
 {
     uint8_t              byte;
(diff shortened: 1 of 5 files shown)


xeioex:
See the discussion here: #871

VadimZhestikov (approved):
Looks good

PR #903, opened 2025-05-02
Tests: fixed js_body_filter.t.

xeioex:
The 1496ed3f commit made visible a problem with the fragile filter tests which depend on the exact sequence of data chunks. The fix is to use perl http server to ensure the order.


VadimZhestikov (approved):
Looks good

PR #904, opened 2025-05-05
HTTP: fixed GCC 15 build with -Wunterminated-string-initialization.

xeioex:
ngx_http_js_module.c:936:22: error: initializer-string for array of ‘char’ truncates NUL terminator
 but destination lacks ‘nonstring’ attribute (16 chars into 15 available) [-Werror=unterminated-string-initialization]
  936 |             .value = "PeriodicSession",
      |                      ^~~~~~~~~~~~~~~~~

### DIFF
--- a/src/njs.h
+++ b/src/njs.h
@@ -187,7 +187,7 @@ struct njs_external_s {
 
     union {
         struct {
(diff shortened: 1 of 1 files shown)


VadimZhestikov (approved):
Looks good

PR #906, opened 2025-05-05
Release 0.9.0

VadimZhestikov:


CHANGES
```diff
@@ -1,3 +1,29 @@
+Changes with njs 0.9.0                                       06 May 2025
+ 
+     nginx modules:
+ 
+     *) Change: refactoring and polishing code for QuickJS fetch module.
```
xeioex, line 5: I think we should remove the block as not very important or remarkable.

CHANGES
```diff
@@ -1,3 +1,29 @@
...
+     *) Feature: refactored working with built-in strings, symbols
+        and small integers.
+        Performance improvements (arewefastyet/benchmarks/v8-v7 benchmark):
+        Richards: +57% (631 → 989)
+        Crypto: +7% (1445 → 1551)
+        RayTrace: +37% (562 → 772)
+        NavierStokes: +20% (2062 → 2465)
+        Overall score: +29% (1014 → 1307)
+
+    *) Change: js_body_filter.t tests.
```
xeioex, line 18: This block is about just tests, not very important for our users.

xeioex (changes requested):
Otherwise looks good.

CHANGES
VadimZhestikov: May be better keep it, as it contains "example for a typical body modification task" ?

xeioex (approved):

PR #910, opened 2025-05-13
Fixed special case of 'for' statement parsing.

VadimZhestikov:
Found by OSS-Fuzz.


xeioex:
The full fix was merged in #1118.

Issue #709, opened 2024-04-25
Add support for statefiles to js_shared_dict_zone.


jo-carter:
For persistently storing the state of the shared dictionary across stop & starts of nginx, and to support loading a pre-defined set of key-value pairs into a shared dictionary without hacks.

There are certain cases where nginx **must** be stopped & started to make configuration changes, such as when adjusting the limit_req_zone's key. This feature would preserve shared_dictionary's values in those cases.

This would work in a similar fashion to [NGINX Plus's Keyval Zone](http://nginx.org/en/docs/http/ngx_http_keyval_module.html#keyval_zone).

PR #915, opened 2025-05-21
Shared dict state file

xeioex:
Modules: added state file for a shared dictionary.

A new optional state parameter is added for js_shared_dict_zone
directive. state parameter specifies a file that keeps the current state
of the shared dict in the JSON format and makes it persistent
across nginx restarts.

This closes https://github.com/nginx/njs/issues/709 feature request on Github.


VadimZhestikov (approved):
Looks good

PR #917, opened 2025-05-23
Add spaces in readme to clarify nginx compile command args with qjs

mcblum:
### Proposed changes

This could be just me, but I originally didn't put a space between the flag and the path. I also thought that `<` was part of the entire command, so that it was supposed to be `-I</opt/quickjs>` which, obviously, doesn't work. This just clarifies and hopefully makes it more obvious that `<` just means "contained within is your var".


xeioex:
Adressed in #925.

Issue #921, opened 2025-05-31
njs_process_script_fuzzer crashing test case in `njs_function_alloc`


gal1ium:
### Describe the bug

Hi! We found a crashing test case when testing with the njs_process_script_fuzzer fuzzing driver.

- [x] The bug is reproducible with the latest version of njs.
- [x] I minimized the code and NGINX configuration to the smallest
possible to reproduce the issue.

### To reproduce

- JS script
```js
var a=[0]
a.forEach(
    function() {
        async function foo(){
            Function`
                });
                let b="";
                b;
                function o(){};
                //`
        }
        foo()
    }
)
```

### Asan Output
```
==3846895==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x00000062fdf2 bp 0xbebebebebebebe4e sp 0x7ffd9424b250 T0)
==3846895==The signal is caused by a READ memory access.
==3846895==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
    #0 0x62fdf2 in njs_function_alloc /src/njs/src/njs_function.c:22:45
    #1 0x634fde in njs_function_constructor /src/njs/src/njs_function.c:1130:16
    #2 0x6329d0 in njs_function_native_call /src/njs/src/njs_function.c:645:11
    #3 0x6329d0 in njs_function_frame_invoke /src/njs/src/njs_function.c:681:16
    #4 0x537eda in njs_vmcode_interpreter /src/njs/src/njs_vmcode.c:1514:15
    #5 0x6331e1 in njs_function_lambda_call /src/njs/src/njs_function.c:608:11
    #6 0x67297a in njs_async_function_frame_invoke /src/njs/src/njs_async.c:28:11
    #7 0x537eda in njs_vmcode_interpreter /src/njs/src/njs_vmcode.c:1514:15
    #8 0x6331e1 in njs_function_lambda_call /src/njs/src/njs_function.c:608:11
    #9 0x61ef6a in njs_function_call /src/njs/src/njs_function.h:167:12
    #10 0x61ef6a in njs_array_iterator_call /src/njs/src/njs_array.c:2117:12
    #11 0x61ef6a in njs_array_handler_for_each /src/njs/src/njs_array.c:2203:16
    #12 0x6703d2 in njs_object_iterate /src/njs/src/njs_iterator.c
    #13 0x61147c in njs_array_prototype_iterator /src/njs/src/njs_array.c:2490:11
[...]

PR #922, opened 2025-06-03
Fixed Function constructor template injection.

xeioex:
The Function constructor uses a `(function(<args>) {<body>})` template to construct new functions. This approach was vulnerable to template injection where malicious code could close the function body early and ignore the trailing brackets using // comments.

The fix is to insert extra newlines into the template, which prevents // comments from commenting out the remaining template.

This fixes issue #921.


VadimZhestikov (approved):
Looks good

VadimZhestikov (changes requested):
.

VadimZhestikov (approved):
Looks good

PR #933, opened 2025-06-17
Fetch: fixed handling of Content-Length header when body is provided.

### LINKED ISSUE #930: Strange error 400 from `ngx.fetch` with body
### Describe the bug

If I want to use `ngx.fetch` to forward request to some other server (aka `proxy_pass`) I'm copying `r.headersIn` and passing `r.requestText` as body. But this method fails.

I've found that I can fix bug by adding `headers.delete('content-length');`. This makes me think that in such config `ngx.fetch` passes two `content-length` headers to server and it fails to parse it.
https://github.com/nginx/njs/blob/master/nginx/ngx_js_fetch.c#L691

### To reproduce

I've created gist with complete nginx.conf and js module:
https://gist.github.com/pyhedgehog/8f3bf0e8373aeb3f75c7b23347a1e578#file-showbug-md

### Expected behavior

Possible solutions:

1. Skip `content-length` from input headers.
2. Skip generating `content-length` if it already in headers (like with `has_host`).
3. Document this behavior.

xeioex:
body value length takes precedence over Content-Length from header list.

    https://fetch.spec.whatwg.org/#http-network-or-cache-fetch

    Let contentLength be httpRequest’s body’s length, if httpRequest’s body
is non-null; otherwise null.

    Let contentLengthHeaderValue be null.

    If httpRequest’s body is null and httpRequest’s method is `POST` or
`PUT`, then set contentLengthHeaderValue to `0`.

    If contentLength is non-null, then set contentLengthHeaderValue to
contentLength, serialized and isomorphic encoded.

    If contentLengthHeaderValue is non-null, then append (`Content-Length`,
contentLengthHeaderValue) to httpRequest’s header list.

This fixes #930 issue in Github.

### DIFF
--- a/nginx/ngx_js_fetch.c
+++ b/nginx/ngx_js_fetch.c
@@ -514,6 +514,7 @@ ngx_js_ext_fetch(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
     ngx_url_t            u;
     ngx_uint_t           i;
     njs_bool_t           has_host;
+    ngx_str_t            method;
     ngx_pool_t          *pool;
     njs_value_t         *init, *value;
     ngx_js_http_t       *http;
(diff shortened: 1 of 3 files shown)


VadimZhestikov (approved):
Looks good

PR #953, opened 2025-08-15
Modules: revert js_set checks introduced in d157f56.

xeioex:
d157f56 introduced, among other things, configure time checks for variables created using js_set directive. It turned out to be too strict. It rejected configurations where js_import is declared at server level or below and not at main level.

This reverts d157f56, but adds a log warning when js_import was not found.

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -1540,6 +1540,9 @@ ngx_http_js_variable_set(ngx_http_request_t *r, ngx_http_variable_value_t *v,
     }
 
     if (rc == NGX_DECLINED) {
+        ngx_log_error(NGX_LOG_WARN, r->connection->log, 0,
+                      "no \"js_import\" directives found for \"js_set\" \"%V\"",
+                      fname);
         v->not_found = 1;
         return NGX_OK;
     }
@@ -7691,14 +7694,10 @@ ngx_http_js_init_conf_vm(ngx_conf_t *cf, ngx_js_loc_conf_t *conf)
 static ngx_int_t
 ngx_http_js_init(ngx_conf_t *cf)
 {
-    ngx_uint_t                  i, found_issue;
-    ngx_js_set_t               *data;
-    ngx_hash_key_t             *key;
-    ngx_http_variable_t        *v;
-    ngx_js_periodic_t          *periodic;
-    ngx_js_loc_conf_t          *jlcf;
-    ngx_js_main_conf_t         *jmcf;
-    ngx_http_core_main_conf_t  *cmcf;
+    ngx_uint_t           i, found_issue;
+    ngx_js_periodic_t   *periodic;
+    ngx_js_loc_conf_t   *jlcf;
+    ngx_js_main_conf_t  *jmcf;
 
     ngx_http_next_header_filter = ngx_http_top_header_filter;
     ngx_http_top_header_filter = ngx_http_js_header_filter;
@@ -7724,21 +7723,6 @@ ngx_http_js_init(ngx_conf_t *cf)
             }
 
             found_issue = 1;
-        }
-
-        cmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_core_module);
-        key = cmcf->variables_keys->keys.elts;
-
-        for (i = 0; i < cmcf->variables_keys->keys.nelts; i++) {
-            v = key[i].value;
-            if (v->get_handler == ngx_http_js_variable_set) {
-                data = (ngx_js_set_t *) v->data;
-                ngx_log_error(NGX_LOG_EMERG, cf->log, 0,
-                              "no \"js_import\" directives found for "
-                              "\"js_set\" \"$%V\" \"%V\" in %s:%ui", &v->name,
-                              &data->fname, data->file_name, data->line);
-                found_issue = 1;
-            }
         }
 
         if (found_issue) {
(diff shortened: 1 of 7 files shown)


nginx/ngx_http_js_module.c
```diff
@@ -1540,6 +1540,9 @@ ngx_http_js_variable_set(ngx_http_request_t *r, ngx_http_variable_value_t *v,
     }
 
     if (rc == NGX_DECLINED) {
+        ngx_log_error(NGX_LOG_WARN, r->connection->log, 0,
```
p-pautov: Should probably be NGX_LOG_ERR just like "js function ... not found".

nginx/ngx_http_js_module.c
```diff
@@ -7691,14 +7694,10 @@ ngx_http_js_init_conf_vm(ngx_conf_t *cf, ngx_js_loc_conf_t *conf)
...
-    ngx_uint_t                  i, found_issue;
-    ngx_js_set_t               *data;
-    ngx_hash_key_t             *key;
-    ngx_http_variable_t        *v;
-    ngx_js_periodic_t          *periodic;
-    ngx_js_loc_conf_t          *jlcf;
-    ngx_js_main_conf_t         *jmcf;
-    ngx_http_core_main_conf_t  *cmcf;
+    ngx_uint_t           i, found_issue;
+    ngx_js_periodic_t   *periodic;
```
p-pautov: Doesn't the same issue apply to periodics?

What location is `jlcf` here from?

nginx/ngx_http_js_module.c
xeioex: I was thinking about it as similar to [WARN in set directive](https://github.com/nginx/nginx/blob/master/src/http/modules/ngx_http_rewrite_module.c#L212).

WARN, because it can be ignored, returning an empty variable value.

But, looking at ERR "js function ... not found" which should be similar to this message, I agree that ERR is less confusing.

nginx/ngx_http_js_module.c
xeioex: >Doesn't the same issue apply to periodics?

yes, you are right


> What location is jlcf here from?

This is location block which is part of a main block. This is done for ease of merging.
But you made me think we can do better and track the correct location block.

nginx/ngx_http_js_module.c
```diff
@@ -1540,6 +1540,9 @@ ngx_http_js_variable_set(ngx_http_request_t *r, ngx_http_variable_value_t *v,
     }
 
     if (rc == NGX_DECLINED) {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "no \"js_import\" directives found for \"js_set\" \"%V\""
+                      " in the currect scope", fname);
```
p-pautov: The `\"js_set\" \"%V\"` part looks a bit confusing, as if a word is missing there. Perhaps it should be `\"js_set\" function \"%V\"`.

Also, "currect" -> "current".

nginx/ngx_http_js_module.c
```diff
@@ -7707,43 +7710,48 @@ ngx_http_js_init(ngx_conf_t *cf)
...
+    cmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_core_module);
+    key = cmcf->variables_keys->keys.elts;
 
-        for (i = 0; i < cmcf->variables_keys->keys.nelts; i++) {
-            v = key[i].value;
-            if (v->get_handler == ngx_http_js_variable_set) {
-                data = (ngx_js_set_t *) v->data;
+    for (i = 0; i < cmcf->variables_keys->keys.nelts; i++) {
+        v = key[i].value;
+        if (v->get_handler == ngx_http_js_variable_set) {
```
p-pautov: TBH, I don't think its worth having all this validation code if runtime check is still required. So I'd suggest to remove config time validation, like it was in previous revision (but for both js_set and periodics). We really need to check variable usage place not the declaration, but Nginx doesn't provide such hooks...

Besides, the js_set change can break some existing code.

nginx/t/js_variables_location.t
```diff
@@ -0,0 +1,96 @@
...
+
+###############################################################################
+
+like(http_get('/foo'), qr/foo_var/, 'foo var');
+like(http_get('/bar'), qr/bar_var/, 'bar var');
+like(http_get('/not_found'), qr/NOT_FOUND:$/, 'not found is empty');
+
+$t->stop();
+
+ok(index($t->read_file('error.log'),
```
p-pautov, line 92: The log message is checked here, but not in stream?

nginx/ngx_http_js_module.c
xeioex: I was unsure about it.

But, yes, I agree that if runtime validation is still require the configure time checks look less useful.

nginx/ngx_http_js_module.c
```diff
@@ -4485,7 +4485,16 @@ ngx_http_js_periodic_handler(ngx_event_t *ev)
...
+    if (rc == NGX_DECLINED) {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "no \"js_import\" directives found for \"js_periodic\""
+                      " handler \"%V\" in the current scope",
+                      &periodic->method);
+        ngx_http_js_periodic_destroy(r, periodic);
+        return;
+    }
+
+    if (rc == NGX_ERROR) {
```
p-pautov: Why not keep `rc != NGX_OK`? That way it's clear we are not expecting anything else here without reading `ngx_http_js_init_vm`.

xeioex: rc == NGX_ERROR - means we had a memory error. I do not think we should log the message about `js_import` in this case.

p-pautov: I'm not suggesting to log a message in `rc != NGX_OK` case. Just keep it as a final check:
```suggestion
    if (rc != NGX_OK) {
```

p-pautov (approved):
Looks good!

VadimZhestikov (approved):
Looks good

Issue #960, opened 2025-09-02
QuickJS support broke in the last week!


mcblum:
### Describe the bug

I don't exactly know what happened, but I went to build our container a few days ago and it just crashed out. Investigating today, I'm seeing an error of `nginx [emerg] invalid value for js_engine "qjs"`. I started throwing things at the wall I mean being a qualified and capable developer and found that if I pinned our install to `972951a4286101776c3a4cee2a33ac3dd4be4ec9` then everything was fine again. I tried pinning QuickJS to its commit a couple weeks ago, but that didn't work, so I'm mostly sure that it's an Njs thing.

It should be reproducible, but our setup is building Open Resty from source, QuickJS from source, and enabling FIPS for everything, so the install file is definitely not your standard one.

- [ x ] The bug is reproducible with the latest version of njs.
- [ x ] I minimized the code and NGINX configuration to the smallest
possible to reproduce the issue.

### To reproduce

I can try to separate all of this stuff out into a script that allows you to reproduce it, but it'll take a second and I'm hoping you "just know" what the issue is :)

### Expected behavior

QuickJS should still work within Njs.

### Your environment

- Latest 
- Open Resty 1.27
- Lua JIT + Headers More
- Hardened Ubuntu with FIPS support

### Additional context

Thank you very much, I sure do love writing Typescript within Nginx!

PR #961, opened 2025-09-02
QuickJS: fix nginx configure with not configured njs module.

VadimZhestikov:
This fixes issue #960 which was introduced in 8259f9a.


src/quickjs_adapted.h
```diff
@@ -0,0 +1,24 @@
...
+    #pragma GCC diagnostic pop
+#else
+    #include <quickjs.h>
+#endif
+
+#ifndef JS_BOOL
+    #define JS_BOOL bool
+#endif
+
+#include <pthread.h>
```
xeioex: It seems we can get rid of it on a common quickjs include. It is only needed for njs_shell.c.

src/quickjs_adapted.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#ifndef __has_warning
+    #define __has_warning(x) 0
+#endif
+
+#if (defined(__GNUC__) && (__GNUC__ >= 8))                                    \
+    || (defined(__clang__) && __has_warning("-Wcast-function-type"))
+    #pragma GCC diagnostic push
+        #pragma GCC diagnostic ignored "-Wcast-function-type"
```
xeioex: style: it should be indented to the same level as previous #pragma.

auto/quickjs
```diff
@@ -15,7 +15,7 @@ if [ $NJS_TRY_QUICKJS = YES ]; then
     njs_feature_run=yes
     njs_feature_incs="$NJS_QUICKJS_DEFAULT_INCS"
     njs_feature_libs="-lquickjs.lto -lm -ldl -lpthread"
-    njs_feature_test="#include <qjs.h>
+    njs_feature_test="#include <quickjs_adapted.h>
```
xeioex: A more common suffix, which is already used for this in njs repo is `compat`.

xeioex (changes requested):
Looks good.
See the remaining fixup patches here https://github.com/xeioex/njs/tree/reenable

VadimZhestikov:
Applied suggested fixups

PR #963, opened 2025-09-11
Fetch keepalive support.

xeioex:


nginx/ngx_js_http.c
```diff
@@ -1528,3 +1574,230 @@ ngx_js_check_header_name(u_char *name, size_t len)
...
+        return NGX_ERROR;
+    }
+
+    ngx_log_debug2(NGX_LOG_DEBUG_EVENT, http->log, 0,
+                   "js http save keepalive connection, "
+                   "saving connection: %p:%d", c, c->fd);
+
+    conf = http->conf;
+
+    if (ngx_queue_empty(&conf->fetch_keepalive_cache)
```
hongzhidao, line 1738: What about moving the creating into configuration phase?

nginx/ngx_js_http.c
```diff
@@ -210,18 +213,32 @@ ngx_js_http_resolve_done(ngx_js_http_t *http)
...
-    if (http->peer.connection != NULL) {
+    if (http->peer.connection == NULL) {
+        return;
+    }
+
+    ngx_log_debug0(NGX_LOG_DEBUG_EVENT, http->log, 0,
+                   "js http close peer");
+
+    if (http->conf->fetch_keepalive > 0 && !http->http_parse.connection_close) {
+        if (ngx_js_http_save_keepalive_connection(http) != NGX_OK) {
```
hongzhidao: It seems we can set the return of `ngx_js_http_save_keepalive_connection()` `void`.
In the function there is also `http->peer.connection = NULL;`.
And the alternative naming `ngx_js_http_free_keepalive_connection()` also looks good.
```
ngx_js_http_get_keepalive_connection()
ngx_js_http_free_keepalive_connection()
vs
ngx_js_http_get_keepalive_connection()
ngx_js_http_save_keepalive_connection()
```
Both are good.

nginx/ngx_js_http.h
```diff
@@ -21,6 +30,7 @@ typedef struct {
     u_char                        *status_text_end;
     ngx_uint_t                     count;
     ngx_flag_t                     chunked;
```
hongzhidao: It seems we can move these fields into `ngx_http_js_t` from `ngx_js_http_parse_t`.
```
    ngx_flag_t                     chunked;
    ngx_flag_t                     connection_close;
    off_t                          content_length_n;
```
What about using `keepalive` instead of `connection_close`, and set its default `1`.

nginx/ngx_js_http.c
xeioex: I want to avoid preallocating this data for large configurations with thousands of locations.
As of now, keep_alive is disable by default. But I plan to enable it after 0.9.3.

nginx/ngx_js_http.c
xeioex: >It seems we can set the return of ngx_js_http_save_keepalive_connection() void.
> In the function there is also http->peer.connection = NULL;.


not sure. after `http->peer.connection = NULL` there is only a success path.
before it the function may return not NGX_OK - in this case we have to close the connection anyway.

xeioex: renamed `ngx_js_http_save_keepalive_connection ()` to `ngx_js_http_free_keepalive_connection ()` to make the code more similar to upstream keepalive module.

nginx/ngx_js_http.h
xeioex: Sounds good. will do.

nginx/ngx_js_http.c
hongzhidao: Makes sense.

nginx/ngx_js_fetch.c
```diff
@@ -682,13 +677,22 @@ ngx_js_ext_fetch(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
...
+        }
+
         njs_chb_append(&http->chain, h[i].key.data, h[i].key.len);
         njs_chb_append_literal(&http->chain, ": ");
         njs_chb_append(&http->chain, h[i].value.data, h[i].value.len);
         njs_chb_append_literal(&http->chain, CRLF);
     }
 
-    njs_chb_append_literal(&http->chain, "Connection: close" CRLF);
+    if (http->conf->fetch_keepalive == 0) {
```
hongzhidao: Or using `!http->keepalive` instead.

nginx/ngx_js_http.h
```diff
@@ -13,15 +13,22 @@
...
 typedef struct ngx_js_http_s  ngx_js_http_t;
 
+typedef struct {
+    ngx_js_loc_conf_t     *conf;
+    ngx_queue_t            queue;
+    ngx_connection_t      *connection;
+
+    socklen_t              socklen;
+    ngx_sockaddr_t         sockaddr;
+} ngx_js_http_keepalive_cache_t;
```
hongzhidao: Can it be moved to `ngx_js_http.c`?

hongzhidao (dismissed):

VadimZhestikov (dismissed):

xeioex:
TODO: remove ip address match when looking for a connection to reuse. Use (scheme, host, port) tuple instead.

PR #970, opened 2025-09-30
Rethinking exit hook.

xeioex:


nginx/ngx_http_js_module.c
```diff
@@ -1701,7 +1701,16 @@ ngx_http_js_cleanup_ctx(void *data)
 
     jlcf = ngx_http_get_module_loc_conf(r, ngx_http_js_module);
 
+    /*
+     * r->pool set to NULL by ngx_http_free_request().
+     * Creating a temporary pool for potential use in njs.on('exit', ...)
+     * handler.
+     */
+    r->pool = ngx_create_pool(128, ctx->log);
```
VadimZhestikov, line 1709: Small note: may be instead of creating of temporary pool, shift ngx_http_free_request() to be executed after exit handler to solve this issue?

VadimZhestikov (approved):
Looks good.

nginx/ngx_http_js_module.c
xeioex: ngx_http_js_cleanup_ctx() itself is executed inside a ngx_destroy_pool() [here](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_request.c#L3847).

```
ngx_http_free_request()
   ...
    pool = r->pool;
    r->pool = NULL;
    ngx_destroy_pool(pool);
        ngx_http_js_cleanup_ctx()
            here may we need r->pool again for njs.on('exit', ...)
```

VadimZhestikov: Ok, it is something to think about later

PR #972, opened 2025-10-02
Exception while loading.

xeioex:


nginx/ngx_js.c
VadimZhestikov, line 581: Should we call njs_vm_destroy(vm) here?

VadimZhestikov: NGX_OK ?

xeioex: yep. Good catch.

Issue #956, opened 2025-08-22
Support HTTP forward proxy in `ngx.fetch()`


route443:
### Is your feature request related to a problem? Please describe

Corporate environments often require some/all egress traffic to pass through a forward proxy. In nginx+njs, subrequests cannot be used for this (nginx oss/plus lacks HTTP CONNECT support), so `ngx.fetch()` is the only outgoing http client available inside njs. Today `ngx.fetch()` has no way to use a connect method.

### Describe the solution you'd like

```js
await ngx.fetch(url, {
  method: 'GET',
  headers,
  body,
  proxy: 'http://user:pass@proxy.example.com:3128',
  proxy_auth: { scheme: 'basic' } // maybe this is unnecessary
});

PR #976, opened 2025-10-11
Fetch: added forward proxy support with HTTPS tunneling.

xeioex:
This patch adds forward proxy support to the ngx.fetch() method in njs, enabling HTTP and HTTPS requests through a proxy server with optional
  authentication.
                                     
  Key Features:                                                                                                                                       
                                                                                                                                                      
  - HTTP proxying: Uses absolute-form URIs when sending requests through proxy
  - HTTPS tunneling: Implements CONNECT method for HTTPS requests through proxy                                                                       
  - Proxy authentication: Supports Basic authentication via Proxy-Authorization header
  - Keepalive support: Works with existing js_fetch_keepalive directive
                                                                                                                                                      
  New Directives:                                                                                                                                     

  - js_fetch_proxy <url> - Configures forward proxy URL
  
  Example nginx.conf:

```nginx.conf
...
  http {
      js_import script.js;

      server {
          listen 8080;
          
          resolver   127.0.0.1:5353;

          # Configure forward proxy
          location /api {
              js_fetch_proxy http://user:pass@proxy.example.com:3128;
              js_content script.fetch_data;
          }
      }
  }
  ```

  JavaScript usage:

```js
  async function fetch_data(r) {
      // Will automatically use configured proxy
      let reply = await ngx.fetch('https://api.example.com/data');
      let body = await reply.text();
      r.return(200, body);
  }
  
 export default {fetch_data};
  
```

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -15,6 +15,8 @@
 typedef struct {
     NGX_JS_COMMON_LOC_CONF;
 
+    ngx_http_complex_value_t  fetch_proxy_cv;
+
     ngx_str_t              content;
     ngx_str_t              header_filter;
     ngx_str_t              body_filter;
@@ -380,6 +382,8 @@ static char *ngx_http_js_content(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf);
 static char *ngx_http_js_shared_dict_zone(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf);
+static char *ngx_http_js_fetch_proxy(ngx_conf_t *cf, ngx_command_t *cmd,
+    void *conf);
 static char *ngx_http_js_body_filter_set(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf);
 static ngx_int_t ngx_http_js_init_conf_vm(ngx_conf_t *cf,
@@ -591,6 +595,13 @@ static ngx_command_t  ngx_http_js_commands[] = {
       ngx_conf_set_msec_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
       offsetof(ngx_http_js_loc_conf_t, fetch_keepalive_timeout),
+      NULL },
+
+    { ngx_string("js_fetch_proxy"),
+      NGX_HTTP_MAIN_CONF|NGX_HTTP_SRV_CONF|NGX_HTTP_LOC_CONF|NGX_CONF_TAKE1,
+      ngx_http_js_fetch_proxy,
+      NGX_HTTP_LOC_CONF_OFFSET,
+      0,
       NULL },
 
       ngx_null_command
@@ -815,6 +826,16 @@ static njs_external_t  ngx_http_js_ext_request[] = {
         .u.property = {
             .handler = ngx_http_js_ext_get_request_body,
             .magic32 = NGX_JS_BUFFER,
+        }
+    },
+
+    {
+        .flags = NJS_EXTERN_PROPERTY,
+        .name.string = njs_str("requestLine"),
+        .enumerable = 1,
+        .u.property = {
+            .handler = ngx_js_ext_string,
+            .magic32 = offsetof(ngx_http_request_t, request_line),
         }
     },
 
@@ -1080,6 +1101,8 @@ static const JSCFunctionListEntry ngx_http_qjs_ext_request[] = {
     JS_CGETSET_DEF("remoteAddress", ngx_http_qjs_ext_remote_address, NULL),
     JS_CGETSET_MAGIC_DEF("requestBuffer", ngx_http_qjs_ext_request_body, NULL,
                          NGX_JS_BUFFER),
+    JS_CGETSET_MAGIC_DEF("requestLine", ngx_http_qjs_ext_string, NULL,
+                         offsetof(ngx_http_request_t, request_line)),
     JS_CGETSET_MAGIC_DEF("requestText", ngx_http_qjs_ext_request_body, NULL,
                          NGX_JS_STRING),
     JS_CGETSET_MAGIC_DEF("responseBuffer", ngx_http_qjs_ext_response_body, NULL,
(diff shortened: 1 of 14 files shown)


nginx/ngx_qjs_fetch.c
```diff
@@ -352,8 +371,16 @@ ngx_qjs_ext_fetch(JSContext *cx, JSValueConst this_val, int argc,
     }
 
     http->naddrs = 1;
-    ngx_memcpy(&http->addr, &u.addrs[0], sizeof(ngx_addr_t));
-    http->addrs = &http->addr;
+
+    if (ngx_js_proxy(http->conf)) {
+        ngx_memcpy(&http->addr, &http->conf->fetch_proxy_url->addrs[0],
+                   sizeof(ngx_addr_t));
+        http->addrs = &http->addr;
```
VadimZhestikov: Same lines 378 and 381 can be moved out of and placed after 'if' statement. See same place in ngx_js_fetch.c

nginx/ngx_qjs_fetch.c
```diff
@@ -237,6 +237,7 @@ ngx_qjs_ext_fetch(JSContext *cx, JSValueConst this_val, int argc,
     ngx_int_t            rc;
     ngx_url_t            u;
     ngx_pool_t          *pool;
+    ngx_str_t           *resolve_host;
```
VadimZhestikov, line 239: Move line 240 before 239. See same place in ngx_js_fetch.c

nginx/ngx_js_http.c
```diff
@@ -24,6 +24,8 @@ typedef struct {
 
 
 #define ngx_js_http_version(major, minor)  ((major) * 1000 + (minor))
+#define NGX_JS_USER_AGENT                                                    \
+    "nginx-js/1.0 (nginx/" NGINX_VERSION " njs/" NJS_VERSION ")"
```
VadimZhestikov: Let's by default do not expose versions, I'd suggest instead use "nginx-js/1.0", or just "nginx-js".
It follows nodejs behavior,  which use "node".

nginx/ngx_js.c
```diff
@@ -3305,6 +3305,85 @@ ngx_js_preload_object(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+    ngx_url_t          *u;
+    ngx_str_t          *value;
+    ngx_str_t           userpass, b64;
+    ngx_js_loc_conf_t  *jscf;
+
+    jscf = conf;
+
+    value = cf->args->elts;
+
+    if (ngx_strncmp(value[1].data, "http://", sizeof("http://") - 1) != 0) {
```
VadimZhestikov: Username and password can both contain ':' and '@', so it seems we need follow url encoding rules?
Or may be better get username and password from separate strings (not from url) to avoid url encoding/decoding complexities?

nginx/ngx_js_http.c
xeioex: Thank you for pointing out. I left it as `nginx-js`.

nginx/ngx_js.c
xeioex: Thank you for pointing out.  I added url decoding + tests.

nginx/t/js_fetch_proxy.t
```diff
@@ -0,0 +1,193 @@
...
+            js_content test.http_fetch;
+        }
+
+        location /http_via_proxy_bad_auth {
+            js_fetch_proxy http://wronguser:wrongpass@127.0.0.1:%%PORT_8081%%;
+            js_content test.http_fetch_status;
+        }
+
+        location /http_via_proxy_encoded {
+            js_fetch_proxy http://user%40domain:p%40ss%3Aword@127.0.0.1:%%PORT_8081%%;
```
VadimZhestikov, line 70: May be better avoid using url format with username and password because it is obsolete and looks confusing for configuration (there is not very obvious set of chars to be url encoded by admin), can we replace it by 2 clear separated strings for username and password?

Something like:

js_fetch_proxy http://127.0.0.1:%%PORT_8081%%  "user_name" "password"

xeioex: @route443  Hi Ivan,

Do you have any opinion/preference/experience here?

route443: Hi @xeioex , I can’t say I prefer the `http://ip:port user pass` format over `http://user:pass@ip:port`. The latter is still widely used, and encoding special chars isn’t much of a problem. It became obsolete mainly because creds could end up in logs and were considered insecure, however, that’s not really relevant in our case. In my opinion, the lack of support for nginx complex values is actually a bigger limitation - it would eliminate this concern altogether and add more flexibility.

xeioex: @route443 

> In my opinion, the lack of support for nginx complex values is actually a bigger limitation - it would eliminate this concern altogether and add more flexibility.

you would like to see a complex value for `js_fetch_proxy` argument?

route443: Undoubtedly, that would provide additional flexibility both in terms of creds management and proxy selection.
For example, if you look at the current OIDC [approach](https://github.com/nginxinc/nginx-openid-connect/blob/24d53f992c81cad1646d9b2f60dd4ff397545a40/openid_connect_configuration.conf#L23), endpoint selection is handled through a `map`. It'd be convenient to manage proxy configs the same way.

Unfortunately, with the current approach, we’d still need to explicitly enable proxy support (by uncommenting the corresponding line for [openid_connect.server_conf](https://github.com/nginxinc/nginx-openid-connect/blob/24d53f992c81cad1646d9b2f60dd4ff397545a40/openid_connect.server_conf#L49)), but it would allow having different proxies/creds for different, say, hosts.

As I suggested in the #956, the greatest flexibility would come from implementing it via ngx.fetch(), though as I understand it, that would be more complex to implement and less transparent for the end admin.
It would also be nice to be able to dynamically enable or disable proxies altogether…

That said, even the current implementation will cover 99% of use cases, since I doubt that:
- anyone uses more than one proxy for the same purpose, or
- anyone intends to split token set retrieval requests between proxy and direct modes.

And even in that case, workarounds could be found.

nginx/ngx_http_js_module.c
```diff
@@ -8016,6 +8039,60 @@ ngx_http_js_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+    return ngx_js_parse_proxy_url(pool, r->connection->log, &value,
+                                  url_out, auth_out);
+}
+
+
+static char *
+ngx_http_js_fetch_proxy(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+    ngx_uint_t                        n;
+    ngx_str_t                        *value;
```
VadimZhestikov, line 8065: Move line 8066 one line up

nginx/ngx_stream_js_module.c
```diff
@@ -3683,6 +3694,62 @@ ngx_stream_js_shared_dict_zone(ngx_conf_t *cf, ngx_command_t *cmd,
...
+
+static char *
+ngx_stream_js_fetch_proxy(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+    ngx_uint_t                         n;
+    ngx_str_t                         *value;
+    ngx_stream_js_srv_conf_t          *jscf;
+    ngx_stream_compile_complex_value_t ccv;
+
+    jscf = conf;
```
VadimZhestikov: Line 3727 can be moved into if statement below

nginx/ngx_stream_js_module.c
```diff
@@ -3683,6 +3694,62 @@ ngx_stream_js_shared_dict_zone(ngx_conf_t *cf, ngx_command_t *cmd,
...
+    return ngx_js_parse_proxy_url(pool, s->connection->log, &value,
+                                  url_out, auth_out);
+}
+
+
+static char *
+ngx_stream_js_fetch_proxy(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+    ngx_uint_t                         n;
+    ngx_str_t                         *value;
```
VadimZhestikov, line 3714: Move line 3723 one line up

nginx/ngx_http_js_module.c
```diff
@@ -8016,6 +8039,60 @@ ngx_http_js_set(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+    ngx_http_js_loc_conf_t           *jlcf;
+    ngx_http_compile_complex_value_t  ccv;
+
+    jlcf = conf;
```
VadimZhestikov: Line 8070 can be moved into if statement below

VadimZhestikov (dismissed):
Couple minor comments, otherwise looks good to me.

VadimZhestikov (approved):
Looks good

PR #982, opened 2025-11-03
Modules: improved handling of results of async handlers.

VadimZhestikov:
Previously, r.setReturnValue() had to be used when returning a value from async js_set handler.

async function hash(r) {
    let hash = await crypto.subtle.digest('SHA-512', r.headersIn.host);
    r.setReturnValue(Buffer.from(hash).toString('hex'));
}

Now r.setReturnValue() is not needed:

async function hash(r) {
    let hash = await crypto.subtle.digest('SHA-512', r.headersIn.host);
    return Buffer.from(hash).toString('hex');
}

Also added promise handling in global qjs code.


VadimZhestikov:
Obsolete

PR #987, opened 2025-11-11
Using printing and parsing library from QuickJS.

xeioex:
- Number.prototype.toString(radix) Improved accuracy for edge cases Reimplemented using njs_dtoa2() with JS_DTOA_FORMAT_FREE | JS_DTOA_EXP_DISABLED
- Number.prototype.toFixed(frac) Reimplemented using njs_dtoa2() with JS_DTOA_FORMAT_FIXED Removed old njs_fixed_dtoa() implementation
- Number.prototype.toPrecision(prec) Reimplemented using njs_dtoa2() with precision format Removed old njs_dtoa_precision() implementation
- Number.prototype.toExponential(frac) Reimplemented using njs_dtoa2() with exponential format Removed old njs_dtoa_exponential() implementation
- parseInt() Simplified parsing implementation Removed custom njs_number_radix_parse() helper
- parseFloat() Simplified parsing implementation

Removed custom njs_number_bin_parse(), njs_number_oct_parse(), njs_number_dec_parse() and njs_strtod.c module Better handling of large numbers and denormal floats and invalid inputs.

+4 test262 tests passed.


src/njs_string.c
```diff
@@ -608,7 +608,8 @@ njs_string_prototype_concat(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
     njs_int_t          ret;
     njs_uint_t         i;
     njs_string_prop_t  string;
-    char               buf[512];
+#define NJS_DTOA_MAX_LEN  njs_length("-1.7976931348623157e+308")
+    char               buf[512], num_buf[NJS_DTOA_MAX_LEN];
```
VadimZhestikov: njs_length(s) returns length without trailing \0

so we need use:

num_buf[NJS_DTOA_MAX_LEN+1]

because njs_dtoa() appends \0 to output.

src/njs_string.c
VadimZhestikov, line 650: Are we safe here after new njs_dtoa append \0 ?

src/njs_string.c
```diff
@@ -724,8 +725,13 @@ njs_string_prototype_concat(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
                     np += sz;
 
                 } else {
-                    sz = njs_dtoa(num, (char *) p);
-                    p += sz;
+                    /*
```
VadimZhestikov: May be better just modify njs_dtoa, so it will not append \0 ? and keep our code here w/o changes?

And provide comment in njs_dtoa,c about how it was imported from quickjs? I make easier propagation fixes from quickjs, if any.

src/njs_json.c
```diff
@@ -751,30 +751,17 @@ static const u_char *
...
 {
-    double        num;
-    njs_int_t     sign;
-    const u_char  *start;
+    double      num;
+    const char  *start, *end;
 
-    sign = 1;
+    start = (const char *) p;
+    num = njs_atod(start, &end, 10, 0);
```
VadimZhestikov: Parsing need to be limited by ctx->end as it was done previously

xeioex:
@VadimZhestikov 

> What about njs_dtoa.c supportability? We have no build time sync with changes in original quickjs/dtoa.c.

I do not expect many changes in the library. We will be able to do it manually from time to time.
In the same way QuickJS-NG is importing patches from the original QuickJS.

src/njs_string.c
```diff
@@ -611,6 +611,7 @@ njs_string_prototype_concat(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
     char               buf[512];
 
 #define NJS_SZ_LAST    64
+#define NJS_DTOA_MAX_LEN  25 /*njs_dtoa_max_len(d, 10, 0, JS_DTOA_FORMAT_FREE)*/
```
VadimZhestikov: style: spaces after  '\/*' and before '\*/'

src/njs_number.c
```diff
@@ -553,13 +427,13 @@ static njs_int_t
...
     njs_uint_t nargs, njs_index_t unused, njs_value_t *retval)
 {
-    u_char       *p;
-    int64_t      frac;
-    double       number;
-    size_t       length, size;
-    njs_int_t    ret, point, prefix, postfix;
-    njs_value_t  *value;
-    u_char       buf[128], buf2[128];
+    int64_t        frac;
```
VadimZhestikov: style: move ''int64_t        frac;" 2 lines down.

src/njs_number.c
```diff
@@ -948,45 +697,21 @@ njs_int_t
 njs_number_parse_int(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
     njs_index_t unused, njs_value_t *retval)
 {
-    double             num;
     int32_t            radix;
+    double             num;
```
VadimZhestikov, line 700: style: move 'int 32_t   radix;' one line down

src/njs_lexer.c
```diff
@@ -813,80 +813,60 @@ njs_lexer_string(njs_lexer_t *lexer, njs_lexer_token_t *token, u_char quote)
...
 
-        /* Binary literal values. */
-
-        if (*p == 'b' || *p == 'B') {
-            p++;
+    /*
+     * njs_atod() does not take the size or end pointer, it relies on the
+     * null-terminated string or non-numerical character to stop parsing.
+     * It is save to call it here because the lexer input is always
+     * null-terminated when compiled or ends with a non-numerical character
```
VadimZhestikov: Better practice do not relay on details of memory allocation mechanics for strings. Currently we overcomplicate possible modifications there, if somebody want it, if any. May be in some situation can be obtained segmentation fault at access to 1st byte of memory out of string.

xeioex: This is a trade-off.

Things to consider
1) `njs_atod_buf()` will require copying unknown length of input data.
2) adding length check to `njs_atod()` will make it slow. Also, it follows standard `strtod()` behaviour.
3) we know it is safe to call `njs_atod()` here, for the reasons given.
4) QuickJS does the same in similar place.

I will make this assumption explicit by adding a comment similar to https://github.com/bellard/quickjs/blob/master/quickjs.h#L826

src/njs_lexer.c
```diff
@@ -813,80 +813,59 @@ njs_lexer_string(njs_lexer_t *lexer, njs_lexer_token_t *token, u_char quote)
...
+        {
+            flags |= JS_ATOD_INT_ONLY;
         }
+    }
 
-        /* Binary literal values. */
-
-        if (*p == 'b' || *p == 'B') {
-            p++;
+    /*
```
VadimZhestikov: All strings are null terminated now, so comment looks obsolete now

VadimZhestikov (approved):
Otherwise looks good.

VadimZhestikov (approved):
good catch

PR #989, opened 2025-11-19
Quickjs ng 0.11.0

xeioex:


nginx/config
```diff
@@ -63,6 +63,21 @@ if [ $NJS_QUICKJS != NO ]; then
         . auto/feature
     fi
 
+    if [ $ngx_found = no ]; then
+        if /bin/sh -c "(pkg-config quickjs-ng --exists)" >> /dev/null 2>&1; then
```
BalkanMadman: Do we need to call another shell here? The following should work as well without unnecessary shell exec.

```suggestion
        if pkg-config quickjs-ng --exists; then
```

BalkanMadman: The `--exists` option doesn't print anything

nginx/config
```diff
@@ -63,6 +63,21 @@ if [ $NJS_QUICKJS != NO ]; then
...
+
+            # pkg-config
+
+            ngx_feature="QuickJS-NG library via pkg-config"
+
+            ngx_feature_path="$NJS_QUICKJS_DEFAULT_INCS"
+            ngx_feature_path="$ngx_feature_path `pkg-config quickjs-ng --cflags | sed -n -e 's#.*-I *\([^ ][^ ]*\).*#\1#p'`"
```
BalkanMadman: Is there a need for such complicated `sed`? The pkg-config's `--cflags-only-I` option to get only include CFLAGS. Then, a simple `sed 's/-I//g'` should remove the `-I` suffixes from all includes, as required by `ngx_feature_path`. Additionally, the backtick command substitution is deprecated in POSIX sh, `$(command --arg1 --arg2 etc)` should be used instead.

The `-e` parameter is redundant when only one expression is supplied to sed.

```suggestion
            ngx_feature_path="$ngx_feature_path $(pkg-config quickjs-ng --cflags-only-I | sed 's/-I//g')"
```

nginx/config
```diff
@@ -63,6 +63,21 @@ if [ $NJS_QUICKJS != NO ]; then
...
+            ngx_feature_libs=`pkg-config quickjs-ng --libs`
```
BalkanMadman: Ditto.

```suggestion
            ngx_feature_libs=$(pkg-config quickjs-ng --libs)
```

auto/quickjs
```diff
@@ -49,6 +49,21 @@ if [ $NJS_TRY_QUICKJS = YES ]; then
         . auto/feature
     fi
 
+    if [ $njs_found = no ]; then
+        if /bin/sh -c "(pkg-config quickjs-ng --exists)" >> $NJS_AUTOCONF_ERR 2>&1; then
```
BalkanMadman: The same as for `nginx/config`.

```suggestion
        if pkg-config quickjs-ng --exists; then
```

BalkanMadman: `pkg-config --exists` doesn't print anything even if the specified .pc module doesn't exist.

auto/quickjs
```diff
@@ -49,6 +49,21 @@ if [ $NJS_TRY_QUICKJS = YES ]; then
...
+
+            # pkg-config
+
+            njs_feature="QuickJS-NG library via pkg-config"
+
+            njs_feature_incs="$NJS_QUICKJS_DEFAULT_INCS"
+            njs_feature_incs="$njs_feature_incs `pkg-config quickjs-ng --cflags | sed -n -e 's#.*-I *\([^ ][^ ]*\).*#\1#p'`"
```
BalkanMadman: ditto as for `nginx/config`.


```suggestion
            njs_feature_incs="$njs_feature_incs $(pkg-config quickjs-ng --cflags-only-I | sed  's/-I//g')"
```

auto/quickjs
```diff
@@ -49,6 +49,21 @@ if [ $NJS_TRY_QUICKJS = YES ]; then
...
+            njs_feature_libs=`pkg-config quickjs-ng --libs`
```
BalkanMadman:
```suggestion
            njs_feature_libs=$(pkg-config quickjs-ng --libs)
```

src/qjs.h
```diff
@@ -150,6 +150,12 @@ static inline JS_BOOL JS_IsNullOrUndefined(JSValueConst v)
 #define qjs_is_array(cx, a) JS_IsArray(cx, a)
 #endif
 
+#ifdef NJS_HAVE_QUICKJS_IS_ERROR_SINGLE_ARG
+#define qjs_is_error(cx, a) JS_IsError(a)
+#else
+#define qjs_is_error(cx, a) JS_IsError(cx, a)
```
BalkanMadman, line 156: ~~Perhaps, parentheses around `cx` and `a`?~~

Ignore this, it should be fine without additional parentheses.

nginx/config
xeioex: `pkg-config` may not be present. `/bin/sh -c ..` allows to silently eat it.
It is, also, taken from nginx [auto](https://github.com/nginx/nginx/blob/master/auto/feature#L55).

nginx/config
xeioex: > Is there a need for such complicated sed? 
> The -e parameter is redundant when only one expression is supplied to sed.

Usually, we are concerned with portability (old systems, different OSs). And linux and *BSD tooling sometime differ in the way the treat options. That is why I prefer to use commands that we already use.

I looked at pkg-config source, and `--cflags-only-I` exists at least since 2012. Old enough to be used.

> The -e parameter is redundant when only one expression is supplied to sed.

that is maybe implementation defined.

Otherwise applied.

auto/quickjs
xeioex: Applied with some changes.

nginx/config
BalkanMadman: what about using the POSIX-specified built-in `command -v` to check that `pkg-config` exists?

```sh
if command -v pkg-config >/dev/null 2>&1; then
```

I believe it's a bit more natural way to check for command existance.

PR #991, opened 2025-11-20
Fix detached buffer

xeioex:


src/njs_typed_array.c
```diff
@@ -2070,6 +2070,11 @@ njs_typed_array_to_chain(njs_vm_t *vm, njs_chb_t *chain,
     uint32_t           i;
     njs_string_prop_t  separator;
 
+    if (njs_slow_path(njs_is_detached_buffer(array->buffer))) {
+        njs_chb_append(chain, "<detached buffer>", 18);
```
VadimZhestikov: I guess it need throw TypeError instead?

xeioex: It is used in 2 places.
In %TypedArray%.prototype.join(), but here detached array is checked before.
In njs.dump(), here we do not want throw an exception, because it is already a debug output.

VadimZhestikov: then this check need to be moved to njs.dump() to avoid unwanted double checking

VadimZhestikov (dismissed):

PR #993, opened 2025-11-25
Respecting user provided CFLAGS.

xeioex:


BalkanMadman:
What I imagine is something like `CFLAGS="${NJS_REQUIRED_CFLAGS} ${CFLAGS:-${NJS_DEFAULT_CFLAGS}}"`

PR #1000, opened 2025-12-12
HTTP: fixed buffer_type inheritance in if blocks.

### LINKED ISSUE #999: Unexpected type of js_body_filter data argument
Hi,

I'm experiencing strange behavior of `js_body_filter` when combined with the conditional `if` block.
Unexpectedly, the conversion of the `data` argument to `String` type stops when `if` conditional evaluates to true.

Below is a minimal example demonstrating the issue:

```

events {
    worker_connections 1024;
}

http {
    js_import lib.js;

    server {
        listen 80;

        location /origin/ {
            return 200 payload\n;
        }

        location /t1/ {
            proxy_pass http://127.0.0.1/origin/;

            # always false
            if ($remote_addr = nonexistent) { set $x 1; }

            js_header_filter lib.headerFilter;
            js_body_filter   lib.bodyFilter;
        }

        location /t2/ {
            proxy_pass http://127.0.0.1/origin/;

            # always true
            if ($remote_addr) { set $x 1; }

            js_header_filter lib.headerFilter;
            js_body_filter   lib.bodyFilter;
        }

        location /t3/ {
            proxy_pass http://127.0.0.1/origin/;

            # always true, empty block
            if ($remote_addr) { }

            js_header_filter lib.headerFilter;
            js_body_filter   lib.bodyFilter;
        }
    }
}
```

Where `lib.js` has the following content:

```
function bodyFilter(r, data, flags) {
    if (flags.last)
        return r.sendBuffer(data, flags);

    r.sendBuffer(`[${data.constructor.name}] ${data}`);
}

function headerFilter(r) {
    delete r.headersOut['Content-Length'];
}

export default { bodyFilter, headerFilter }
```

The first result meets expectations:

```
% curl http://127.0.0.1/t1/
[String] payload
```

But the second and third ones are not:

```
% curl http://127.0.0.1/t2/
[Buffer] payload

% curl http://127.0.0.1/t3/
[Buffer] payload
```

The case is reproducible with the latest packages from official repo on Ubuntu 22.04.
Thank you!
```

xeioex:
Previously, when js_body_filter was used inside an if block that evaluated to true, the data parameter received Buffer type instead of the expected String type. This happened because buffer_type field in ngx_http_js_loc_conf_t was not properly initialized, causing the configuration merge to fail when nginx created a new location context for if blocks.

This fixes #999 issue on Github.

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -8260,6 +8260,8 @@ ngx_http_js_create_loc_conf(ngx_conf_t *cf)
     conf->ssl_verify = NGX_CONF_UNSET;
     conf->ssl_verify_depth = NGX_CONF_UNSET;
 #endif
+    conf->buffer_type = NGX_CONF_UNSET_UINT;
+
     return conf;
 }
 
--- a/nginx/t/js_body_filter_if.t
+++ b/nginx/t/js_body_filter_if.t
@@ -58,6 +58,24 @@ http {
 
             proxy_pass http://127.0.0.1:8081/source;
         }
+
+        location /type_check_true {
+            if ($arg_check) {
+                set $dummy 1;
+            }
+
+            js_body_filter test.type_check;
+            proxy_pass http://127.0.0.1:8081/backend;
+        }
+
+        location /type_check_false {
+            if ($arg_nonexistent = dummy) {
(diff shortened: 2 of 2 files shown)


VadimZhestikov (approved):
Looks good

PR #1018, opened 2026-01-30
Improved Error.stack.

xeioex:
This makes

```js
try {
} catch(e) {
  log `oops: ${e.stack}`; 
}
```

e.stack value more consistent and efficient.

```
Make Error.stack faster.

Previously, error.stack reported full names for native function (for
example Array.prototype.map -> map), but it was achieved by iteration
through a global object which is slow.

The fix is to report only function name at hand, this loses a bit
of verbosity but make it ~100 times faster.

make benchmark
before
        ...
	exception.stack: 38.421µs, 26027 times/s
	exception.native.stack: 226.711µs, 4410 times/s
after
        ...
	exception.stack: 1.239µs, 807356 times/s
	exception.native.stack: 2.419µs, 413339 times/s
```

### DIFF
--- a/src/test/njs_unit_test.c
+++ b/src/test/njs_unit_test.c
@@ -21403,7 +21403,7 @@ static njs_unit_test_t  njs_shell_test[] =
       njs_str("9") },
 
     { njs_str("var e = Error(); e.name = {}; e" ENTER),
-      njs_str("[object Object]") },
+      njs_str("[object Object]\n    at main (:1)\n") },
 
     { njs_str("var a = []; Object.defineProperty(a, 'b', {enumerable: true, get: Object}); a" ENTER),
       njs_str("[\n b: '[Getter]'\n]") },
@@ -21411,12 +21411,12 @@ static njs_unit_test_t  njs_shell_test[] =
     { njs_str("var e = Error()" ENTER
               "Object.defineProperty(e, 'message', { configurable: true, set: Object })" ENTER
               "delete e.message; e" ENTER),
-      njs_str("Error") },
+      njs_str("Error\n    at main (:1)\n") },
 
     { njs_str("var e = Error()" ENTER
               "Object.defineProperty(e, 'message', { configurable: true, get(){ return 'foo'} })" ENTER
               "e" ENTER),
-      njs_str("Error: foo") },
+      njs_str("Error: foo\n    at main (:1)\n") },
 
     { njs_str("function f() {};" ENTER
               "Object.defineProperty(f, 'name', { get() {void(0)} })" ENTER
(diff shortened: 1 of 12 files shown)


test/shell_test_njs.exp
```diff
@@ -67,7 +67,7 @@ njs_test {
 # Backtraces for external objects
 njs_test {
     {"console.info(console.a.a)\r\n"
-     "console.info(console.a.a)\r\nThrown:\r\nTypeError:*at console.info (native)"}
+     "console.info(console.a.a)\r\nThrown:\r\nTypeError:*at info (native)"}
```
VadimZhestikov, line 70: Side note: this test was and is misleading, because we should not have console.info on the stack.

src/test/njs_unit_test.c
```diff
@@ -21514,6 +21514,9 @@ static njs_unit_test_t  njs_shell_test[] =
 
 static njs_unit_test_t  njs_backtraces_test[] =
 {
+    { njs_str("var e = new Error(); e[0] = 1"),
+      njs_str("1") },
```
VadimZhestikov, line 21518: This test passed OK before this fix. What is tested here?

src/test/njs_unit_test.c
```diff
@@ -21670,24 +21673,34 @@ static njs_unit_test_t  njs_backtraces_test[] =
...
               "    at main (:1)\n") },
 
+    { njs_str("var p = new Promise((_, reject) => { reject(new Error('oops'));});"
+              "p.catch((e) => { $r.retval(e.stack) });"),
+      njs_str("Error: oops\n"
+              "    at <anonymous> (:1)\n"
+              "    at Promise (native)\n"
+              "    at main (:1)\n") },
+
+    { njs_str("var e = new Error('oops'); e.stack = 123; e.stack"),
```
VadimZhestikov, line 21686: This test passed OK before this fix. What is tested here?

src/test/njs_unit_test.c
```diff
@@ -21716,7 +21729,7 @@ static njs_unit_test_t  njs_backtraces_test[] =
 
     { njs_str("Math\n.min(1,\na)"),
       njs_str("ReferenceError: \"a\" is not defined\n"
-              "    at Math.min (native)\n"
+              "    at min (native)\n"
```
VadimZhestikov, line 21732: Side note: this test is misleading, it should be no min in stack. Probably this is some common issue for native stuff?

src/test/njs_unit_test.c
xeioex: I had an issue with setting values to an error object in one of the implementation.
I will not hurt to leave it, I think.

src/test/njs_unit_test.c
xeioex: I had an issue with setting values to an error object in one of the implementation.
I will not hurt to leave it, I think.

src/test/njs_unit_test.c
xeioex: agree, but it looks like a separate issue. I will look into it later.

PR #1030, opened 2026-03-02
Fixed string offset map corruption in scope values hash.

xeioex:
The issue was introduced in e7caa46d (0.9.5).  When compile-time UTF-8 string constants were copied into the values hash in njs_scope_value_index(), the string data layout was calculated incorrectly: the map offset did not account for the null terminator added in e7caa46d, the "size" variable was overwritten corrupting the subsequent memcpy, and the offset map was never initialized to zero.

This caused SEGV/SIGBUS crashes for any multi-byte UTF-8 string constant with more than 32 characters when accessing a character at index >= 32 (e.g. via .replace() or bracket notation).  The bug only manifested when the string byte size was 4-byte aligned, as otherwise alignment padding absorbed the missing byte.

The fix factors out njs_string_data_size() and njs_string_data_init() helpers shared by njs_string_alloc() and njs_scope_value_index(), eliminating the duplicated layout logic that caused the divergence.

### DIFF
--- a/src/njs_scope.c
+++ b/src/njs_scope.c
@@ -158,6 +158,9 @@ njs_scope_value_index(njs_vm_t *vm, const njs_value_t *src, njs_uint_t runtime,
     njs_object_prop_t    *pr;
     njs_flathsh_query_t  fhq;
 
+    /* Suppress "may be used uninitialized" warning. */
+    length = 0;
+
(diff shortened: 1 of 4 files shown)


VadimZhestikov (approved):
Looks good

PR #1055, opened 2026-05-13
Fetch: fixed heap buffer overflow in proxy URL credentials.

xeioex:
The destination buffers for the decoded user and password in ngx_js_parse_proxy_url() were a fixed 128 bytes, while the encoded input length was bounded only by the URL length.  Since ngx_unescape_uri() writes at most one byte per input byte, raw credentials longer than 128 bytes overflowed the buffer; the length check ran only after the decode.

The fix is to size the destination buffer based on the encoded input length.

The bug appeared in dea8318 (0.9.4).


VadimZhestikov (approved):
Looks good

PR #1056, opened 2026-05-14
Coverity issues.

xeioex:


nginx/ngx_http_js_module.c
```diff
@@ -4114,9 +4114,6 @@ ngx_http_js_ext_request_form_for_each(njs_vm_t *vm, njs_value_t *args,
     }
 
     this_arg = njs_arg(args, nargs, 2);
-    if (this_arg == NULL) {
```
VadimZhestikov, line 4117: A brief comment in the commit (or inline) citing why njs_arg() cannot be NULL (i.e., it returns &njs_value_undefined for out-of-bounds) would help future readers avoid re-adding this guard.

VadimZhestikov:
One question for commit 3: Is there a test case covering Error.stack output for bound functions? If not, adding one to the njs test suite would lock in the correct behavior and prevent regression.

VadimZhestikov (approved):
2 minor notes, otherwise looks good

nginx/ngx_http_js_module.c
xeioex: `val = njs_arg(args, nargs, n)` is a very common pattern in njs code

xeioex:
>One question for commit 3: Is there a test case covering Error.stack output for bound functions? If not, adding one to the njs test suite would lock in the correct behavior and prevent regression.

https://github.com/nginx/njs/blob/master/src/test/njs_unit_test.c#L22264 covers this.

PR #1057, opened 2026-05-14
HTTP: fixed r.return() with body in js_access.

xeioex:
Previously, when a js_access handler called r.return(status, body), ngx_http_send_response() sent the response but returned NGX_OK.  The access handler then returned NGX_OK to the phase engine, which treated the access check as allowed and continued to the content phase.

As a result, a denied request could still reach proxy_pass or another content handler after the response had already been sent.

Now the access handler finalizes the request if a response was already sent during js_access execution.  This keeps r.return() behavior in js_content unchanged, while making r.return(status, body) terminal in js_access.

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -174,6 +174,8 @@ typedef struct {
 
 
 static ngx_int_t ngx_http_js_access_handler(ngx_http_request_t *r);
+static ngx_int_t ngx_http_js_access_finalize(ngx_http_request_t *r,
+    ngx_http_js_ctx_t *ctx);
 static void ngx_http_js_access_write_event_handler(ngx_http_request_t *r);
 static ngx_int_t ngx_http_js_content_handler(ngx_http_request_t *r);
 static void ngx_http_js_content_event_handler(ngx_http_request_t *r);
@@ -1540,7 +1542,7 @@ ngx_http_js_access_handler(ngx_http_request_t *r)
             return NGX_HTTP_INTERNAL_SERVER_ERROR;
         }
 
-        return ctx->status;
+        return ngx_http_js_access_finalize(r, ctx);
     }
 
     ctx->status = NGX_OK;
@@ -1594,6 +1596,18 @@ done:
         return NGX_AGAIN;
     }
 
+    return ngx_http_js_access_finalize(r, ctx);
+}
+
+
+static ngx_int_t
+ngx_http_js_access_finalize(ngx_http_request_t *r, ngx_http_js_ctx_t *ctx)
+{
+    if (r->header_sent) {
+        ngx_http_finalize_request(r, NGX_OK);
+        return NGX_DONE;
+    }
+
     return ctx->status;
 }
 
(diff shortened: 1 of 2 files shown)


nginx/ngx_http_js_module.c
```diff
@@ -1594,24 +1596,44 @@ ngx_http_js_access_handler(ngx_http_request_t *r)
...
 
+    return ngx_http_js_access_finalize(r, ctx);
+}
+
+
+static ngx_int_t
+ngx_http_js_access_finalize(ngx_http_request_t *r, ngx_http_js_ctx_t *ctx)
+{
+    if (r->header_sent) {
+        ngx_http_finalize_request(r, NGX_OK);
```
VadimZhestikov, line 1617: Is it safe to call this while the response body write may still be in progress (e.g. large async body not fully flushed)? Or does nginx's buffering ensure this is always safe?

VadimZhestikov (dismissed):
Looks good

nginx/ngx_http_js_module.c
xeioex: Yes.  ngx_http_send_response() does not have to flush the whole body synchronously.  If output remains pending, nginx output filters keep it in r->out and set r->buffered/c->buffered/r->postponed.  Calling [ngx_http_finalize_request](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_request.c#L2816)(r, NGX_OK) is the normal nginx handoff for this state: finalize_request() sees the buffered output, installs ngx_http_writer(), and the [writer continues calling](https://github.com/nginx/nginx/blob/master/src/http/ngx_http_request.c#L3076) ngx_http_output_filter(r, NULL) until the body is drained.

nginx/ngx_http_js_module.c
```diff
@@ -1594,24 +1598,52 @@ ngx_http_js_access_handler(ngx_http_request_t *r)
...
+        ngx_http_finalize_request(r, rc);
+        return NGX_DONE;
+    }
+
+    if (r->header_sent) {
+        ngx_http_finalize_request(r, NGX_OK);
+        return NGX_DONE;
+    }
+
     return ctx->status;
```
VadimZhestikov, line 1621: Is it missing guard?
  Redirect attempted when header_sent is also true

  In ngx_http_js_access_finalize, the redirect check comes before the header_sent check:
```
  if (ctx->redirect_uri.len) {           // checked first
      rc = ngx_http_js_internal_redirect(r, ctx);
      ...
  }
  if (r->header_sent) { ... }            // checked second
```
  If pathological JS code calls both r.return(403, 'body') and r.internalRedirect('/foo'), the redirect
  would be attempted after headers were already sent — which nginx's redirect functions are not designed
  to handle. A defensive check at the top of the redirect branch would make this more robust:
```
  if (ctx->redirect_uri.len) {
      if (r->header_sent) {
          /* log warning: redirect after response sent — ignored */
          ngx_http_finalize_request(r, NGX_OK);
          return NGX_DONE;
      }
      rc = ngx_http_js_internal_redirect(r, ctx);
      ...
  }
```
  This is a low-priority concern (it requires user code to do two contradictory things), but it's worth
  considering for defensive robustness.

VadimZhestikov (dismissed):
Minor question, otherwise looks good

nginx/ngx_http_js_module.c
xeioex: Agree, added a  warning.

VadimZhestikov (approved):
Looks good

PR #1060, opened 2026-05-16
Fixed call argument value snapshotting.

### LINKED ISSUE #1059: Evaluation of arguments has changed
Hi,
After upgrade from 0.9.1 to 0.9.8 I noticed that the order of operations has changed.
Please have a look at below minimal test case:
```

events {
    worker_connections 1024;
}

http {
    js_import just.js;

    server {
        location / {
            js_content just.test;
        }
    }
}
```

Where `just.js` contains:
```
function test(r) {
    let a = 1;

    r.return(200, `result:${foo(a, a = 2)}\n`);
}

function foo(x, y) {
    return `${x}:${y}`;
}

export default { test }
```

Previously the result was:
```
% curl 127.1
result:1:2
```

But now, with 0.9.8, both arguments are the same from called function perspective:
```
% curl 127.1
result:2:2
```

I'm not sure if the behavior was fixed or broken, but it was obviously changed :) and that caused part of the code to suddenly stop working.

Thanks for your help!

xeioex:
Since fd5e523f (0.9.7), njs has evaluated all call argument expressions before emitting the frame and PUT_ARG instructions, this fixed await expressions in call arguments.  This preserved argument side effects before callee validation, but left PUT_ARG reading argument values only after later argument expressions had already run.

As a result, an earlier argument backed by a mutable variable slot could observe a later mutation of the same slot.  For example, f(a, a = 2) passed 2 as both arguments instead of preserving the first value as 1. The same issue applied to later argument effects such as getters.

The fix is to preserve arguments using temporary registers where appropriate.

This fixes #1059 issue on Github.

### DIFF
--- a/src/njs_generator.c
+++ b/src/njs_generator.c
@@ -470,6 +470,15 @@ static njs_int_t njs_generate_call_argument_expressions(njs_vm_t *vm,
     njs_generator_t *generator, njs_parser_node_t *node);
 static njs_int_t njs_generate_call_argument_expressions_after(njs_vm_t *vm,
     njs_generator_t *generator, njs_parser_node_t *node);
+static njs_bool_t njs_generate_argument_aliases_mutable_slot(
+    njs_parser_node_t *node);
+static njs_bool_t njs_generate_argument_index_is_mutable(njs_index_t index);
+static njs_bool_t njs_generate_argument_uses_index(njs_parser_node_t *node,
+    njs_index_t index);
+static njs_bool_t njs_generate_argument_may_observe_effect(
+    njs_vm_t *vm, njs_parser_node_t *node);
+static njs_bool_t njs_generate_argument_suffix_may_observe_effect(
+    njs_vm_t *vm, njs_parser_node_t *node);
 static njs_uint_t njs_generate_call_nargs(njs_parser_node_t *node);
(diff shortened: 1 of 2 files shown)


VadimZhestikov:
1. Observation: njs_generate_argument_uses_index has dead recursive tree-walk logic (children are always
   NULL in current call sites). A comment explaining that this function is called only on leaf nodes would
   prevent future confusion — or the child checks could be removed with a njs_assert(node->left == NULL &&
   node->right == NULL).
   
  2. Suggestion: The node->left->index == index pre-check before recursion in
  njs_generate_argument_uses_index would silently fail to detect variable use inside compound subtrees if
  the function's call conditions ever change. Consider replacing with an unconditional recursive call
  (which node->temporary would likely prevent from being reached in practice anyway).

Both are minor.

VadimZhestikov (approved):
Looks good

xeioex:
>Observation: njs_generate_argument_uses_index has dead recursive tree-walk logic (children are always
NULL in current call sites)

The recursive walk is not dead.  This is easily discoverable with simple assert statement + current tests.

It is reached by compound argument expressions whose result aliases a mutable slot, for example assignment expressions `f(a = b, b = 2)` through `left`, and comma expressions `f(a, (a = 2, a), a = 3)` through right.

PR #1061, opened 2026-05-16
Modules: added jsVarNames() method.

xeioex:
The method returns names of variables declared with js_var.  An optional string argument filters returned names by prefix.

This lets JavaScript code discover a configuration-defined variable schema without duplicating every variable name in the script.  For example:

    js_var $extract_name;
    js_var $extract_arguments_workspace;

    function handler(r) {
        r.jsVarNames("extract_").forEach(function(name) {
            r.variables[name] = extract(name.slice(8));
        });
    }

The same API is available as s.jsVarNames() in stream.  Variables created by js_set or by the core modules are not returned.

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -366,6 +366,9 @@ static njs_int_t ngx_http_js_ext_variables(njs_vm_t *vm,
 static njs_int_t ngx_http_js_periodic_session_variables(njs_vm_t *vm,
     njs_object_prop_t *prop, uint32_t atom_id, njs_value_t *value,
     njs_value_t *setval, njs_value_t *retval);
+static njs_int_t ngx_http_js_ext_js_var_names(njs_vm_t *vm,
+    njs_value_t *args, njs_uint_t nargs, njs_index_t unused,
+    njs_value_t *retval);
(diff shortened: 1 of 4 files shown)


VadimZhestikov:
Minor suggestions:

---
  Stream module implementations

  Fully symmetric to the HTTP implementations — ngx_stream_session_t, ngx_stream_core_main_conf_t,
  ngx_stream_js_variable_var used in place of their HTTP counterparts. Logic is identical.

  Minor inconsistency between the two QJS variants — n = 0 is initialised at the declaration site in the
  HTTP version but near its first use in the stream version:
```
  // HTTP QJS:
      n = 0;
      cmcf = ngx_http_get_module_main_conf(r, ngx_http_core_module);
      v = cmcf->variables.elts;
```
```
  // Stream QJS:
      cmcf = ngx_stream_get_module_main_conf(s, ngx_stream_core_module);
      v = cmcf->variables.elts;
      n = 0;
```
  Not a bug, but worth harmonizing.

---

  Tests:

  1. jsVarNames(null) — both implementations would throw TypeError (null is not undefined, not a string),
  but this isn't tested.
  2. jsVarNames(undefined) explicitly passed — both implementations treat it the same as no argument
  (returns all), but not tested explicitly. Low priority.

VadimZhestikov (dismissed):
Looks good

VadimZhestikov (approved):
Looks good

PR #1062, opened 2026-05-16
Fixed allocator mismatch in drain/drop.

xeioex:
njs_chb_destroy() frees chain nodes through chain->free() and guards against a NULL free callback.  njs_chb_drain() and the tail-freeing path of njs_chb_drop() called njs_mp_free() directly, which is wrong for chains initialized with NJS_CHB_CTX_INIT() where chain->free is js_free(), and unsafe for NGX_CHB_CTX_INIT() chains where chain->free is NULL.

Both paths now route through chain->free() with the same NULL guard as njs_chb_destroy().


VadimZhestikov (approved):
Looks good

PR #1066, opened 2026-05-22
Hardening fixes.

xeioex:
The build issue on QuickJS is known, and [reported](https://github.com/bellard/quickjs/issues/515) to upstream.

### DIFF
--- a/src/njs_object.c
+++ b/src/njs_object.c
@@ -1136,9 +1136,10 @@ njs_object_own_enumerate_object(njs_vm_t *vm, const njs_object_t *object,
 
         ret = njs_get_own_ordered_keys(vm, object, parent, items_sorted, flags);
         if (ret != NJS_OK) {
-            return NJS_ERROR;
+            goto fail;
         }
 
+        entry = NULL;
         njs_set_object(&value, (njs_object_t *) object);
 
         for (i = 0; i< items_sorted->length; i++) {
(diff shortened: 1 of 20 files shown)


src/qjs_buffer.c
VadimZhestikov: return JS_EXCEPTION?

VadimZhestikov (dismissed):
Looks good, except for external issues related to changes in bellard/quickjs

src/njs_object.c
```diff
@@ -1163,7 +1164,11 @@ njs_object_own_enumerate_object(njs_vm_t *vm, const njs_object_t *object,
 
             ret = njs_array_add(vm, items, &retval);
             if (njs_slow_path(ret != NJS_OK)) {
-                return NJS_ERROR;
+                if (njs_object_enum_kind(flags) == NJS_ENUM_BOTH) {
```
VadimZhestikov: entry is allocated whenever flags != NJS_ENUM_VALUES (i.e., for both KEYS and BOTH modes), but the cleanup guard is == NJS_ENUM_BOTH. For KEYS-only mode,  entry would leak on njs_array_add failure. The condition should be != NJS_ENUM_VALUES to mirror the allocation condition.

VadimZhestikov (approved):
Looks good

PR #1067, opened 2026-05-24
Add ngx_js_form_fuzzer for HTTP form body parsing

kuranikaran:
njs 0.9.9 added nginx/ngx_js_form.c, an ~860-line multipart and urlencoded form body parser exposed through the r.readRequestForm() API. The existing njs_process_script_fuzzer only covers the JS engine script path and doesn't reach this parser.
Since this code handles untrusted HTTP request bodies (Content-Type parsing, boundary matching, percent-decoding, Content-Disposition header extraction, quoted-string unescaping), it's a high-value fuzzing target.
This PR adds a dedicated fuzz harness that drives ngx_js_parse_form() directly with both urlencoded and multipart inputs through minimal nginx core stubs (pool, array, string helpers), avoiding the need for a full nginx build.
What's included:

fuzz/ngx_js_form_fuzzer.cc — harness source
fuzz/stubs/ — minimal nginx type/function stubs (ngx_pool, ngx_array, ngx_str)
fuzz/seed_corpus/ — 8 seed inputs covering normal, truncated, oversized-boundary, quoted-filename, and empty-body paths
fuzz/ngx_js_form_fuzzer.dict — fuzzer dictionary

This is being coordinated with a corresponding [oss-fuzz integration PR](https://github.com/google/oss-fuzz/pull/15533) to wire up the new fuzzer in Google's continuous fuzzing infrastructure.
Note: This PR intentionally does not modify auto/make or the build system. The oss-fuzz integration ([google/oss-fuzz#15533](https://github.com/google/oss-fuzz/pull/15533)) injects the build target dynamically. If you'd prefer the target wired into auto/make natively, happy to add that.


kuranikaran:

kuranikaran:
Hi @xeioex  - this adds a fuzzer for the new form body parser in ngx_js_form.c. Would appreciate your review when you get a chance. The corresponding oss-fuzz integration is at google/oss-fuzz#15533.

xeioex:
Hi @kuranikaran,

thank you for the patch.

I do not mind adding more fuzzing targets in principle, but  I am not sure about the current approach which adds a lot of shim code. Also ParseSeedText() produces relatively structured input to generate  genuinely unexpected inputs.

Given that it adds complexity and you found no issues with the code (thank you for that!) I do not think it worth the effort to fuzz this function continually (unlike the general JS file input).

kuranikaran:
Thanks @xeioex, that's fair - the shim surface is real maintenance cost and you're right that a target that surfaced nothing isn't worth carrying continually. I'll close this out. Appreciate the point on ParseSeedText; I'll revisit the input generation locally to make sure the clean run wasn't just shallow coverage. Closing the corresponding oss-fuzz PR too so nothing dangles.

PR #1069, opened 2026-06-03
Exceptions overhaul

xeioex:
Cleaning up of exception classes for QuickJS/njs integration code + small fixes found.  The goal is not to rewrite
all messages, but to make exception classes predictable while keeping real bug fixes visible for review and backports.

- Wrong receiver and wrong user-visible API usage are `TypeError`.
- Numeric/status bounds violations are `RangeError`.
- Host/runtime failures remain `InternalError`.
- Allocation-only failures use memory errors / `OutOfMemory`.
- Correctness fixes are split from pure exception-class retyping when the split is small and clear.

## Correctness fixes split out

- Fetch QuickJS conversion handling:
  - preserve exceptions already raised by QuickJS conversions;
  - add `JS_HasException()` / `JS_ThrowOutOfMemory()` guards where `ngx_qjs_string()` can fail silently on pool allocation;
  - use `JS_GetOpaque()` instead of `JS_GetOpaque2()` where class mismatch is a normal miss, avoiding stale pending exceptions;
  - free the headers init value before throwing for non-object `Headers`.
- HTTP njs handlers:
  - set pending exceptions before returning `NJS_ERROR` from response output paths such as `sendHeader()`, `send()`, and `finish()`.
- Stream variables:
  - clear `vv->valid` and set memory error when stream variable storage allocation fails.

## Exception-class changes

- Fetch, HTTP, Stream, XML, and common helper code now use `TypeError` for receiver/API misuse.
- Fetch/HTTP/Stream status and code range checks use `RangeError`.
- XML parse failures use `SyntaxError`.
- Internal nginx/output/body collection failures remain `InternalError`.
- `ngx_js_ext_log()` / `ngx_qjs_ext_log()` missing external receiver/context is treated as `TypeError`, matching the wrong-receiver policy.
- Stream `s.done()` while filtering is treated as `TypeError`, matching other wrong-phase user API calls.

## Deliberate non-goals

- Shared dictionary exception alignment is not included as it requires additional work
- Existing generic conversion messages in HTTP/Stream handlers are mostly kept to avoid unnecessary user-visible churn.
- No extra minor receiver tests were added just to exercise every retyped branch.

### DIFF
--- a/external/njs_xml_module.c
+++ b/external/njs_xml_module.c
@@ -2022,7 +2022,7 @@ njs_xml_error(njs_vm_t *vm, njs_xml_doc_t *current, const char *fmt, ...)
                         err->int2);
     }
 
-    njs_vm_error(vm, "%*s", p - errstr, errstr);
+    njs_vm_syntax_error(vm, "%*s", p - errstr, errstr);
 }
 
 
--- a/external/qjs_fs_module.c
+++ b/external/qjs_fs_module.c
@@ -2399,7 +2399,7 @@ qjs_fs_stats_get_own_property(JSContext *cx, JSPropertyDescriptor *pdesc,
 
     st = JS_GetOpaque2(cx, obj, QJS_CORE_CLASS_ID_FS_STATS);
     if (st == NULL) {
-        (void) JS_ThrowInternalError(cx, "\"this\" is not a Stats object");
+        (void) JS_ThrowTypeError(cx, "\"this\" is not a Stats object");
         return -1;
     }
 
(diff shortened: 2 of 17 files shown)


VadimZhestikov (approved):
Two suggestions, otherwise looks good:

  1. QJS/njs asymmetry (commit 3): ngx_qjs_request_ctor uses JS_ThrowInternalError for "input is not string or a Request
  object" where the njs path uses njs_vm_type_error. Should be JS_ThrowTypeError.
  2. Minor classification inconsistency (commit 5): "failed to convert uri arg" / "failed to convert options.method"
  etc. are labeled TypeError but the conversion failure at those points is almost always OOM (the value type was
  verified earlier). Removing these explicit messages (as done elsewhere in the PR) would be cleaner and let njs set the
  right error type.

xeioex:
> QJS/njs asymmetry (commit 3): ngx_qjs_request_ctor uses JS_ThrowInternalError for "input is not string or a Request object" where the njs path uses njs_vm_type_error. Should be JS_ThrowTypeError.

Is fixed by 16f8534f.

PR #1070, opened 2026-06-04
QuickJS: fix consumed value cleanup

xeioex:
QuickJS property setter and define APIs consume the JSValue argument on both success and failure.  This applies to JS_SetProperty(), JS_SetPropertyStr(), JS_SetPropertyUint32(), JS_DefinePropertyValue(), JS_DefinePropertyValueStr(), and JS_DefinePropertyValueUint32().

Do not free values after passing them to these APIs, including failure paths in shared cleanup labels.  Target objects and containers are not consumed by these calls and still need normal cleanup unless ownership has been transferred by inserting them into another object.

Also fix adjacent cleanup paths that leaked target objects after a setter consumed the value being attached.

Reported by R4mbb of KRsecurity.

### DIFF
--- a/external/qjs_fs_module.c
+++ b/external/qjs_fs_module.c
@@ -1074,12 +1074,13 @@ done:
 
 
 static JSValue
-qjs_fs_dirent_create(JSContext *cx, JSValueConst name, struct dirent *entry)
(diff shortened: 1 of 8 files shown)


VadimZhestikov (approved):
Looks good

PR #1072, opened 2026-06-10
Fix OOM in staging/sm/String/replace-math Tests262.

xeioex:


VadimZhestikov (approved):
Looks good.

One minor item to address (up to you): Move the NJS_CHB_ERR_NONE/NOMEM/OVERFLOW #define constants out of the typedef struct body to before the struct, which is the conventional placement in C and consistent with the rest of the njs header style.

xeioex:
> One minor item to address (up to you): Move the NJS_CHB_ERR_NONE/NOMEM/OVERFLOW #define constants out of the typedef struct body to before the struct, which is the conventional placement in C and consistent with the rest of the njs header style.

Actually, we already have code patterns like this, for example

https://github.com/nginx/njs/blob/master/nginx/ngx_http_js_module.c#L109

but there are more.

PR #1074, opened 2026-06-12
Misc fixes.

xeioex:


nginx/ngx_qjs_fetch.c
```diff
@@ -315,6 +318,7 @@ ngx_qjs_ext_fetch(JSContext *cx, JSValueConst this_val, int argc,
 
         if (!JS_IsUndefined(value)) {
             http->ssl_verify = JS_ToBool(cx, value);
```
VadimZhestikov, line 320: JS_ToBool can return -1 if an exception is thrown (e.g., from a custom
  [Symbol.toPrimitive] on the passed value). The return value is not checked. Compare with places where JS_ToInt64 failures are explicitly caught. 

For consistency and robustness:
```
  rc = JS_ToBool(cx, value);
  JS_FreeValue(cx, value);
  if (rc < 0) { goto fail; }
  http->ssl_verify = rc;
```
  In practice, verify is almost always passed a plain boolean literal, so this is
  low-severity. But it's worth noting given how carefully the adjacent buffer_size and max_response_body_size code is handled.

VadimZhestikov (approved):
One minor note, otherwise looks good.

nginx/ngx_qjs_fetch.c
xeioex: Agree, in principle, but we have several place with JS_ToBool(), probably they need a dedicated pass.

PR #1075, opened 2026-06-12
HTTP: fix segfault reading a request header with no cache slot

### LINKED ISSUE #1071: Worker segfault in r.headersIn[name] for a header with no nginx handler (e.g. Proxy-Connection)
### Describe the bug

A worker process crashes with SIGSEGV while a JS handler reads a request header via `r.headersIn[<name>]`. The fault is a `memcpy` in `njs_chb_append0` from a near-null source pointer (`0x14`). In the symbolized cores the header being read was `Proxy-Connection`. The header's real parsed entry in `r->headers_in.headers` is intact (`value="Keep-Alive"`, `next=NULL`), but the getter dereferences a *different*, garbage `ngx_table_elt_t`. Reading the same headers via `r.rawHeadersIn` instead of `r.headersIn[name]` makes the crash go away.

- [x] The bug is reproducible with the latest version of njs. *(njs 0.9.9; crash is intermittent — see "To reproduce".)*
- [ ] I minimized the code and NGINX configuration to the smallest possible to reproduce the issue. *(Config/JS below are minimal, but the crash is intermittent and we could not produce a guaranteed single-request repro — the production pattern crashes reliably in aggregate. Details under Additional context.)*

### To reproduce

- JS script (suspected minimal trigger):
```js
function handler(r) {
    r.return(200, String(r.headersIn['Proxy-Connection']));
}
export default { handler };
```

- NGINX configuration:
```nginx
events {}
http {
    js_import test.js;
    server {
        listen 8080;
        location / { js_content test.handler; }
    }
}
```

- NGINX logs (kernel-logged worker crash; hostname/addresses redacted):
```
kernel: nginx[<pid>]: segfault at 14 ip <addr> sp <addr> error 4 in libc.so.6[...]
```

Symbolized backtrace (njs 0.9.9, frame pointers enabled):
```
#0  __memcpy_avx_unaligned
#1  njs_chb_append0 (msg=0x14, len=56)                    src/njs_chb.c:38   (#2 = same fn, :28)
#3  ngx_http_js_header_generic (ph=..., flags=0)          nginx/ngx_http_js_module.c:5507
#4  ngx_http_js_header_in (name="proxy-connection")       nginx/ngx_http_js_module.c:5177
#5  ngx_http_js_ext_header_in                             nginx/ngx_http_js_module.c:4433
#6  njs_value_property (atom_id=787)                      src/njs_value.c:1103
#7  njs_vmcode_interpreter                                src/njs_vmcode.c:306
#8  njs_function_lambda_call                              src/njs_function.c:597
#9  njs_async_function_frame_invoke                       src/njs_async.c:28
#10 ngx_engine_njs_call                                   nginx/ngx_js.c:813
#11 ngx_http_js_content_event_handler                     nginx/ngx_http_js_module.c:1718
#12 ngx_http_read_client_request_body
#13 ngx_http_js_content_handler                           nginx/ngx_http_js_module.c:1675
```

- Output of the `nginx -V` command (build-path / file-prefix-map flags redacted):
```
nginx version: nginx/1.31.1
built by gcc 15.2.0 (Ubuntu 15.2.0-16ubuntu1)
built with OpenSSL 3.5.5 27 Jan 2026
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --modules-path=/usr/lib/nginx/modules
  --with-compat --with-threads --with-file-aio
  --with-http_ssl_module --with-http_v2_module --with-http_v3_module
  --with-http_realip_module --with-http_auth_request_module --with-http_sub_module
  --with-stream --with-stream_ssl_module --with-stream_ssl_preread_module --with-stream_realip_module
  --with-mail --with-mail_ssl_module
  --with-cc-opt='-g -O2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -flto=auto -ffat-lto-objects -fstack-protector-strong'
  --with-ld-opt='-flto=auto -ffat-lto-objects -Wl,-z,relro -Wl,-z,now -pie'
  [build-path / file-prefix-map flags redacted]
```

- Exact steps to reproduce the behavior:
  1. Build/run nginx with the dynamic `ngx_http_js_module` from njs 0.9.9 and the config above.
  2. Send HTTP/1.1 requests carrying a `Proxy-Connection` header:
     `curl -H 'Proxy-Connection: keep-alive' http://127.0.0.1:8080/`
  3. Under sustained traffic, a worker eventually segfaults. (See caveat below — a single request may not fault.)

### Expected behavior

`r.headersIn['Proxy-Connection']` should return the header value (or `undefined`), never crash the worker. Reading any request header should be memory-safe regardless of whether nginx has a dedicated handler for that header name.

### Your environment

- Version of njs: **0.9.9**
- Version of NGINX: **1.31.1** (dynamic `ngx_http_js_module`)
- List of other enabled nginx modules: http_v2, http_v3, http_ssl, http_realip, http_auth_request, http_sub, stream (+ ssl / ssl_preread / realip), mail (+ ssl)
- OS: Ubuntu, x86-64; glibc (`__memcpy_avx_unaligned`)

### Additional context

**Observed (facts, from a symbolized core):**
- `name` passed to the getter = `"proxy-connection"`; `r->http_version == 1001` (HTTP/1.1).
- The real parsed entry for the header is intact: `value="Keep-Alive"` (len 10), `next=NULL` (request had 5 headers, all `next=NULL`).
- The `*ph` the getter actually dereferenced was **not** that entry:
  ```
  { hash  = 0x6294400b3d40,   // a pointer into r's own pool region (r = 0x6294400b41e0); not a valid hash
    key   = {len=5,  data=0x0},
    value = {len=56, data=0x14},
    next  = 0x5 }
  ```
  `njs_chb_append0` then `memcpy`'d from `value.data = 0x14`.
- The `msg=0x14, len=56` signature recurred across multiple cores (including an earlier nginx 1.31.0 build, same njs 0.9.9).

**Production code that crashed, and the fix that stopped it (observed):**
```js
// BEFORE — crashed intermittently inside an async js_content handler
for (var h in r.headersIn) {
    headers[h] = r.headersIn[h];   // <-- faults on the r.headersIn[name] read
}

// AFTER — stable, no recurrence since deploy
let raw = r.rawHeadersIn;
for (let i = 0; i < raw.length; i++) {
    headers[raw[i][0]] = raw[i][1];
}
```
The fault is on the `r.headersIn[name]` read itself; `r.rawHeadersIn` walks the parsed list directly and never enters the per-name getter that faults.

**Hypothesis (unverified):** `Proxy-Connection` has no handler in nginx's `ngx_http_headers_in[]`, so no cached `r->headers_in.*` slot; the per-name lookup appears to resolve `ph` to uninitialized/incorrect memory and follow it. Uninitialized contents → usually harmless, occasionally a non-null invalid pointer → crash, which matches the intermittency.

**Reproducibility caveat:** the crash is intermittent in production (many requests between occurrences). We could not produce a deterministic single-request repro; the minimal config above is the suspected trigger, while the `for..in` + async shape is what reliably crashes in aggregate. We have not confirmed whether the `async` context or `for..in` (vs. direct access) is required.

**Impact:** remotely triggerable worker crash (a client-supplied header takes down the worker for any njs config reading it / iterating `r.headersIn`); the worker respawns, so it degrades to dropped in-flight requests rather than RCE / info-leak.

Disclaimer: Claude Code summarized the more technical parts (issue/context/backtrace/hypothesis) and helped identify the root from the core dump, as I am not an expert on that.

xeioex:
Reading a request header via r.headersIn[name] could crash the worker with SIGSEGV.  The header in question was "Proxy-Connection", but any header that nginx registers without a dedicated ngx_table_elt_t * slot was affected.

The getter mapped an arbitrary header name through ngx_hash_find() and treated (char *) &r->headers_in + hh->offset as a pointer to a cache slot.  This only holds for headers whose parser stores into such a slot.  "Proxy-Connection" is registered in ngx_http_headers_in[] with offset 0 and a handler that stores nothing, so the getter aliased the ngx_list_t at the start of ngx_http_headers_in_t and followed its elts pointer as a bogus ngx_table_elt_t.

The fix is to walk the parsed header list by name, as it is always safe. The single-value and semicolon-join flags moved into the per-name header table, and the now-unused slot pointer argument and its branch were dropped from the generic getter.

The slot lookup that is removed was only an optimization for the names in ngx_http_headers_in[]; for any other header it missed the hash and walked the list anyway, so dropping it is not a performance regression.

This closes #1071 issue on Github.

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -389,8 +389,8 @@ static njs_int_t ngx_http_js_header_out_special(njs_vm_t *vm,
     ngx_http_request_t *r, njs_str_t *v, njs_value_t *setval,
     njs_value_t *retval, ngx_table_elt_t **hh);
 static njs_int_t ngx_http_js_header_generic(njs_vm_t *vm,
-    ngx_http_request_t *r, ngx_list_t *headers, ngx_table_elt_t **ph,
-    unsigned flags, njs_str_t *name, njs_value_t *retval);
+    ngx_http_request_t *r, ngx_list_t *headers, unsigned flags,
+    njs_str_t *name, njs_value_t *retval);
 #endif
 static njs_int_t ngx_http_js_content_encoding(njs_vm_t *vm,
     ngx_http_request_t *r, unsigned flags, njs_str_t *name,
@@ -4393,20 +4393,24 @@ static njs_int_t
 ngx_http_js_ext_header_in(njs_vm_t *vm, njs_object_prop_t *prop, uint32_t atom_id,
     njs_value_t *value, njs_value_t *unused, njs_value_t *retval)
 {
-    unsigned             flags;
-    njs_int_t            rc;
-    njs_str_t            name, *h;
-    ngx_http_request_t  *r;
+    unsigned               flags;
+    njs_int_t              rc;
+    njs_str_t              name;
+    ngx_http_request_t    *r;
+    ngx_http_js_header_t  *h;
 
-    static njs_str_t single_headers_in[] = {
-        njs_str("Content-Type"),
-        njs_str("ETag"),
-        njs_str("From"),
-        njs_str("Max-Forwards"),
-        njs_str("Referer"),
-        njs_str("Proxy-Authorization"),
-        njs_str("User-Agent"),
-        njs_str(""),
+    static ngx_http_js_header_t  headers_in[] = {
+#define header(name, fl)  { njs_str(name), fl, 0 }
+        header("Content-Type",        NJS_HEADER_SINGLE),
+        header("ETag",                NJS_HEADER_SINGLE),
+        header("From",                NJS_HEADER_SINGLE),
+        header("Max-Forwards",        NJS_HEADER_SINGLE),
+        header("Referer",             NJS_HEADER_SINGLE),
+        header("Proxy-Authorization", NJS_HEADER_SINGLE),
+        header("User-Agent",          NJS_HEADER_SINGLE),
+        header("Cookie",              NJS_HEADER_SEMICOLON),
+        header("",                    0),
+#undef header
     };
 
     r = njs_vm_external(vm, ngx_http_js_request_proto_id, value);
@@ -4429,11 +4433,11 @@ ngx_http_js_ext_header_in(njs_vm_t *vm, njs_object_prop_t *prop, uint32_t atom_i
 
     flags = 0;
 
-    for (h = single_headers_in; h->length > 0; h++) {
-        if (h->length == name.length
-            && ngx_strncasecmp(h->start, name.start, name.length) == 0)
+    for (h = headers_in; h->name.len > 0; h++) {
+        if (h->name.len == name.length
+            && ngx_strncasecmp(h->name.data, name.start, name.length) == 0)
         {
-            flags |= NJS_HEADER_SINGLE;
+            flags = h->flags;
             break;
         }
     }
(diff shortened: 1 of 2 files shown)


VadimZhestikov (approved):
Looks good


Small note about:

ngx_http_js_header_t struct
```
typedef struct {
    ngx_str_t   name;
    unsigned    flags;
    uintptr_t   handler;
} ngx_http_js_header_t;
```
The handler field is set to 0 in all current entries. Given its uintptr_t type, this is likely a placeholder for function pointers used in response header tables (where per-header special handling exists)? For the headers_in[] use case here, handler = 0 is consistent and correct.

PR #1079, opened 2026-06-15
Fix use-after-free in Array.prototype.sort()

xeioex:
Previously, njs_sort_indexed_properties() cached the fast array backing store pointer (array->start) once and then iterates.  For a hole it performed a generic property lookup that walks the prototype chain and may invoke a user-defined getter.  A getter that grows the same array triggers njs_array_expand(), which reallocates the backing store and frees the old buffer, leaving the cached pointer dangling.  The remaining iterations then read freed memory, and the stale values are sorted and written back, exposing freed heap contents to script.

Reported by Sangsoo Jeong (78ResearchLab).

### DIFF
--- a/src/test/njs_unit_test.c
+++ b/src/test/njs_unit_test.c
@@ -7970,6 +7970,16 @@ static njs_unit_test_t  njs_test[] =
               "njs.dump([undefined, 3, /*hole*/, 2, undefined, /*hole*/, 1].sort())"),
       njs_str("[1,2,3,4,undefined,undefined,<empty>]") },
 
+    /* A prototype getter for a hole reallocates the array being sorted. */
+
+    { njs_str("Object.defineProperty(Array.prototype, 1, {configurable: true,"
+              "  get() { for (var i = 0; i < 1024; i++) { this.push(i); }"
+              "          return 5; }});"
+              "var a = [3]; a[2] = 7; a.length = 3;"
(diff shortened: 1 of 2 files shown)


src/test/njs_unit_test.c
```diff
@@ -7970,6 +7970,16 @@ static njs_unit_test_t  njs_test[] =
...
+    /* A prototype getter for a hole reallocates the array being sorted. */
+
+    { njs_str("Object.defineProperty(Array.prototype, 1, {configurable: true,"
+              "  get() { for (var i = 0; i < 1024; i++) { this.push(i); }"
+              "          return 5; }});"
+              "var a = [3]; a[2] = 7; a.length = 3;"
+              "a.sort(function(x, y) { return x - y; });"
+              "a[0] === 3 && a[1] === 5 && a[2] === 7"),
+      njs_str("true") },
+
```
VadimZhestikov, line 7982: Optional improvement would be adding:

 delete Array.prototype[1]

cleanup to the test case for better test isolation.

VadimZhestikov (approved):
Looks good

src/test/njs_unit_test.c
xeioex: As njs does clone of the context, it is safe to modify global prototypes in tests.
No need to clean it up.

PR #1084, opened 2026-06-24
QuickJS: add missing btoa() and atob()

xeioex:
The functions implement the WHATWG btoa()/atob() global functions for the QuickJS engine, mirroring the built-in njs engine.

The narrower njs-only btoa()/atob() unit tests are dropped in favor of test/btoa.t.js, which exercises both engines.


src/qjs_buffer.c
```diff
@@ -2313,6 +2313,170 @@ qjs_base64url_decode_length(JSContext *ctx, const njs_str_t *src)
...
+    b = bytes;
+
+    njs_utf8_decode_init(&ctx);
+
+    while (p < end) {
+        cp = njs_utf8_decode(&ctx, &p, end);
+
+        if (cp > 0xff) {
+            js_free(cx, bytes);
+            return JS_ThrowTypeError(cx, "invalid character (>= U+00FF)");
```
VadimZhestikov: The check is cp > 0xff, meaning U+00FF (255) is valid — only U+0100 and above are rejected. The message ">= U+00FF" implies U+00FF itself is invalid, which is incorrect.

  The test file hardcodes the same wrong message:
  exception: 'TypeError: invalid character (>= U+00FF)'

src/qjs_buffer.c
```diff
@@ -2313,6 +2313,170 @@ qjs_base64url_decode_length(JSContext *ctx, const njs_str_t *src)
...
+
+    tmp = js_malloc(cx, str.length);
+    if (tmp == NULL) {
+        return JS_ThrowOutOfMemory(cx);
+    }
+
+    p = tmp;
+
+    for (i = 0; i < str.length; i++) {
+        if (str.start[i] == ' ') {
```
VadimZhestikov: The WHATWG "forgiving-base64 decode" algorithm strips all ASCII whitespace: U+0009 TAB, U+000A LF, U+000C FF, U+000D  CR, and U+0020 SPACE. The implementation strips only SPACE.

  The test file documents this as "Forgiving-base64 ignores spaces" and tests only space characters — no TAB, LF, CR, or FF cases. If this is a deliberate deviation from the spec (to mirror the existing njs engine behaviour), it should be  noted; otherwise the stripping condition needs something like:

  if (str.start[i] == ' ' || str.start[i] == '\t' ||
      str.start[i] == '\n' || str.start[i] == '\r' ||
      str.start[i] == '\f')

src/qjs.h
```diff
@@ -131,6 +131,8 @@ JSValue qjs_promise_result(JSContext *cx, JSValue result);
 JSValue qjs_string_hex(JSContext *cx, const njs_str_t *src);
 JSValue qjs_string_base64(JSContext *cx, const njs_str_t *src);
 JSValue qjs_string_base64url(JSContext *cx, const njs_str_t *src);
+JSValue qjs_string_btoa(JSContext *cx, njs_str_t *src);
+JSValue qjs_string_atob(JSContext *cx, njs_str_t *src);
```
VadimZhestikov: New signatures:
```
  JSValue qjs_string_btoa(JSContext *cx, njs_str_t *src);
  JSValue qjs_string_atob(JSContext *cx, njs_str_t *src);
```
  Existing signatures for the adjacent functions:
```
  JSValue qjs_string_hex(JSContext *cx, const njs_str_t *src);
  JSValue qjs_string_base64(JSContext *cx, const njs_str_t *src);
  JSValue qjs_string_base64url(JSContext *cx, const njs_str_t *src);
```
  Neither qjs_string_btoa nor qjs_string_atob modifies *src -- btoa only reads src->length and src->start; atob copies it  immediately (str = *src). Both should be const njs_str_t * to match the established convention.

src/qjs_buffer.c
```diff
@@ -2313,6 +2313,170 @@ qjs_base64url_decode_length(JSContext *ctx, const njs_str_t *src)
...
+        pad = 4 - (len % 4);
+        len += pad;
+    }
+
+    len = qjs_base64_decoded_length(len, pad);
+
+    NJS_CHB_CTX_INIT(&chain, cx);
+
+    dst = njs_chb_reserve(&chain, len * 2);
+    if (dst == NULL) {
```
VadimZhestikov: dst is assigned only to check for NULL. After the check it is never read; all writes go through qjs_chb_write_byte_as_utf8(&chain, ...) which uses njs_chb_current(chain). The u_char *dst declaration can be removed and the reservation written as:
```
  if (njs_chb_reserve(&chain, len * 2) == NULL) {
```

VadimZhestikov (dismissed):
Few notes, otherwise looks good

xeioex:
@VadimZhestikov 

Applied fixes and implemented " algorithm strips all ASCII whitespace", and removed extra tmp allocation from original njs `atob()`. The same changes applied to qjs version.

VadimZhestikov (dismissed):
Looks good

xeioex:
@VadimZhestikov  Fixed assert() for size == 0 in js_malloc_rt() in QuickJS, for empty string output.

VadimZhestikov (approved):
Looks good

PR #1086, opened 2026-07-02
Webcrypto: fix buffer overflow when exporting large keys to JWK

xeioex:
njs_export_base64url_bignum() and qjs_export_base64url_bignum() wrote BN_num_bytes() bytes into a fixed 512-byte stack buffer without a bound check.  An RSA key with a modulus larger than 4096 bits (over 512 bytes), generated or imported via SPKI/PKCS8, overflowed the buffer during exportKey("jwk").

While here, add exception throwing on a parallel import part in both engines.

Reported by Vaibhav Rajput.


VadimZhestikov (approved):
Looks good

PR #1087, opened 2026-07-03
QuickJS: fix unhandled rejection tracking in reused contexts

xeioex:
ngx_qjs_destroy() unregisters the host promise rejection tracker on the runtime, but ngx_qjs_clone() did not re-register it when a context was popped from the reuse queue.  Every request served by a reused context then silently dropped unhandled rejections instead of reporting them, which also defeated any caller relying on ctx->rejected_promises.


VadimZhestikov (approved):
Looks good

PR #1106, opened 2026-08-12
Buffer: fix use-after-free on detach during argument coercion

basavaraj-sm05:
### Proposed changes

Several Buffer.prototype methods, in both engines, fetch the underlying ArrayBuffer's raw data pointer before finishing the coercion of their own offset/length/value arguments. In QuickJS (qjs_buffer.c) this affects the read/write integer and float accessors, write(), copy(), compare(), indexOf(), and toString(); in njs (njs_buffer.c) it affects the shared range helper behind compare()/equals() and the source side of copy(). Any of those arguments can be an object with a valueOf, and the callback it triggers can detach the very buffer whose pointer was already captured, via ArrayBuffer.prototype.transfer() or, in a test context, $262.detachArrayBuffer(). The subsequent bounds check and memory access then run against that stale pointer: a straightforward heap-use-after-free in QuickJS, since transfer() frees the backing store, and a read through a NULL-based address in njs, since its own detach only clears the data pointer. Both are reachable from ordinary script, for example `buf.compare(other, {valueOf(){other.buffer.transfer(); return 0}})`. I found the QuickJS side first, and since njs's own read/write-int functions already re-fetch the buffer after coercion, went looking for the same gap in njs's remaining Buffer methods and found it in the compare/copy range helper too. The fix in both files is the same: resolve every user-controllable argument first, and only look up the buffer pointer immediately before the bounds check and the actual read or write, so a detach mid-conversion now surfaces as the ordinary "detached buffer" error instead of touching freed or invalid memory. I added a regression suite to buffer.t.js that detaches the buffer from inside a valueOf during each affected call and asserts a TypeError is thrown.

### Checklist

### DIFF
--- a/src/njs_buffer.c
+++ b/src/njs_buffer.c
@@ -675,43 +675,43 @@ njs_value_buffer_get(njs_vm_t *vm, njs_value_t *value, njs_str_t *dst)
 
 
 static njs_int_t
-njs_buffer_array_range(njs_vm_t *vm, njs_typed_array_t *array,
-    const njs_value_t *start, const njs_value_t *end, const char *name,
-    uint8_t **out_start, uint8_t **out_end)
+njs_buffer_range_index(njs_vm_t *vm, const njs_value_t *value, uint64_t *out)
 {
-    uint64_t            num_start, num_end;
-    njs_int_t           ret;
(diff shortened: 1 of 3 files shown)


basavaraj-sm05:

PR #1107, opened 2026-08-12
Fetch: validating headers in Headers.set()

VadimZhestikov:
### Proposed changes

`Headers.set()` validates its arguments only when the header does not already
exist. In that case it falls through to `ngx_js_headers_append()`, which trims
optional whitespace and then checks the name with `ngx_js_check_header_name()`,
the value with `ngx_js_check_header_value()`, and the immutable guard.

When the header is already present, `set()` takes a different path: it finds the
existing entry, overwrites `h[i].value` in place and unlinks the duplicates, so
none of those checks run. As a result a value that `append()` rejects is
accepted by `set()`:

```js
var h = new Headers();
h.append('X-A', 'good');
h.append('X-B', 'a\r\nInjected: 1');   // TypeError: invalid header value
h.set('X-A', 'b\r\nInjected: 2');      // accepted
```

The immutable guard is skipped on the same path, so an existing header on an
immutable Headers object can be overwritten.

This moves the checks ahead of the lookup so both paths behave identically. Both
engines have the same structure and both are changed.

After the change the example above throws for `set()` as well, `set()` with an
ordinary value still works, and optional whitespace is trimmed as it is by
`append()`.

### Checklist

- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/nginx/ngx_js_fetch.c
+++ b/nginx/ngx_js_fetch.c
@@ -2073,6 +2073,23 @@ ngx_headers_js_ext_set(njs_vm_t *vm, njs_value_t *args, njs_uint_t nargs,
         return NJS_ERROR;
     }
 
+    ngx_js_http_trim_ows(&value.start, &value.length);
+
+    if (ngx_js_check_header_name(name.start, name.length) != NGX_OK) {
+        njs_vm_type_error(vm, "invalid header name");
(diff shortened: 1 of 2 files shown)


xeioex:
See #1108. I do not amend it here because no original changes left.

PR #1108, opened 2026-08-13
Headers set validation

xeioex:
This is the extended fix for issue reported in https://github.com/nginx/njs/pull/1107

### DIFF
--- a/nginx/ngx_js_fetch.c
+++ b/nginx/ngx_js_fetch.c
@@ -55,6 +55,8 @@ static ngx_int_t ngx_js_fetch_append_headers(ngx_js_http_t *http,
     ngx_js_headers_t *headers, u_char *name, size_t len, u_char *value,
     size_t vlen);
 static void ngx_js_fetch_process_done(ngx_js_http_t *http);
+static njs_int_t ngx_js_headers_result(njs_vm_t *vm,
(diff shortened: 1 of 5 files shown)


VadimZhestikov (dismissed):
Looks good

A small follow-up simplification you might fold in, if you like it:
https://github.com/VadimZhestikov/njs/commit/a554b9fc

Two things on top of this PR, both no-op refactors:

- The per-engine `ngx_js_headers_update()` / `ngx_qjs_headers_update()`
  wrappers only forward to `ngx_js_headers_modify()` and translate the
  result code. `append()` and `set()` can call `ngx_js_headers_modify()`
  and `*_headers_result()` directly, so the wrappers are dropped.
- `ngx_js_headers_modify()` took `value`/`vlen` by pointer, which reads
  like an output parameter, but no caller uses the trimmed value back
  (it's stored inside the function). Passing them by value removes the
  indirection.

No functional change; the fetch tests pass on both njs and QuickJS. The
commit sits right on top of the current PR tip (b2ed5cd), so it applies
clean. Feel free to squash it in or ignore.

xeioex:
Applied.

VadimZhestikov (approved):
Looks good

PR #1111, opened 2026-08-13
Reqest headers validation

xeioex:
This is more extended fix for problem described in #1109.

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -536,6 +536,7 @@ static void *ngx_http_js_create_loc_conf(ngx_conf_t *cf);
 static char *ngx_http_js_merge_loc_conf(ngx_conf_t *cf, void *parent,
     void *child);
 
+static ngx_int_t ngx_http_js_status_is_redirect(ngx_int_t status);
 static ngx_int_t ngx_http_js_parse_unsafe_uri(ngx_http_request_t *r,
     njs_str_t *uri, njs_str_t *args);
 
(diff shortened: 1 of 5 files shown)


VadimZhestikov (approved):
Looks good.

Reviewed and tested -- this supersedes #1109, which is now closed.

It catches two things #1109 missed:

- **Header names.** #1109 validated values only, so `r.headersOut['X\r\nInjected: 1'] = 'v'`
  still split the response. Validating where the name is resolved for a store
  covers `Object.defineProperty()` too.
- **The `r.return()` redirect target**, which reaches Location without ever being
  checked as a header value -- it was only checked as a body, where control
  characters are legal.

The `ngx_js_header_value()` / `ngx_qjs_header_value()` wrappers are also a better
shape than what #1109 did. Converting and validating in one call means a store
path cannot pick up the conversion without the check, which is what went wrong
here in the first place.

Coverage looks right: ten call sites across both engines and both nginx version
branches. The one remaining raw `ngx_js_string(vm, setval, ...)` is in
`ngx_http_js_request_variables()`, which is not a header path -- correctly left
alone.

Tested on both engines, `js_engine qjs` and the default:

    value CRLF            -> invalid header value
    name CRLF             -> invalid header name
    Object.defineProperty -> invalid header name
    r.return(302, ...)    -> invalid header value

with no injected header appearing in any response, and benign traffic unchanged:
plain values, `Location`, and `Set-Cookie` arrays all still work.

One thing that is not yours, but worth knowing: `js_shared_dict.t` tests 35 and
37 ("foo keys after expire", "no of items in foo after expire") failed in one of
two full-suite runs on this branch. They pass 3 out of 3 when the file is run
alone, and this branch does not touch shared dict code, so it looks like an
expiry test that is timing-sensitive under full-suite load -- the extra tests
here shift the timing. Not attributable to this change, just flagging it.

PR #1112, opened 2026-08-14
Pre js context reuse fixes hardening

xeioex:
The first set of robustness/hardening fixes for QuickJS with `js_context_reuse 0;`

### DIFF
--- a/nginx/ngx_http_js_module.c
+++ b/nginx/ngx_http_js_module.c
@@ -2183,10 +2183,8 @@ ngx_http_js_init_vm(ngx_http_request_t *r, njs_int_t proto_id)
 static void
 ngx_http_js_cleanup_ctx(void *data)
 {
-    ngx_http_request_t      *r;
-    ngx_http_js_loc_conf_t  *jlcf;
-
-    ngx_http_js_ctx_t        *ctx = data;
+    ngx_http_request_t  *r;
+    ngx_http_js_ctx_t   *ctx = data;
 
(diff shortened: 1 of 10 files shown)


VadimZhestikov (dismissed):
Looks good. Two non-blocking notes.

  **Accessor liveness.** Agreed this is finalization only, and nothing here
  regresses accessor safety. Worth noting the remaining step is cheap:
  `object->data` is written once and never cleared, and all fourteen
  `ngx_qjs_fetch_object_data()` call sites already handle a NULL return. Clearing
  it when the pool goes away would make a retained `Request`/`Response` degrade
  safely, with no consumer changes.

  **Alignment.** The nit is real:

      ngx_qjs_clear_events:   *event col 24,  *node col 23
      ngx_njs_clear_events:   *event col 23,  *node col 23

  `ngx_qjs_event_t` is one character longer than `ngx_js_event_t`, so the five
  spaces carried over from the twin push `*event` out by one. Four spaces fixes it.

  **Tested.** The four new tests pass (29 tests), and the full `js_*.t` suite
  passes (40 files, 387 tests). Note three of the four skip without `--with-stream`
  and rewrite -- easy to get a green run that did not execute them.

  Also confirmed the reworked finalizers never dereference `object->data`, which is
  what "pool-independent" rests on.

xeioex:
@VadimZhestikov 

I rebased and applied style correction. I left Accessor liveness for a future changes as it requires js_context_reuse > 0.

VadimZhestikov (approved):
Looks good

PR #1115, opened 2026-08-16
Fix compat with quickjs ng 0.16

xeioex:


src/qjs.c
```diff
@@ -252,7 +252,12 @@ qjs_new_context(JSRuntime *rt, qjs_module_t **addons)
     JS_AddIntrinsicMapSet(ctx);
     JS_AddIntrinsicTypedArrays(ctx);
     JS_AddIntrinsicPromise(ctx);
-#ifdef NJS_HAVE_QUICKJS_ADD_INTRINSIC_BIG_INT
+    /*
+     * Since QuickJS-NG 0.16.0 BigInt is a part of the base objects
+     * intrinsic, registering it again leaks the previous constructor.
+     */
+#if defined(NJS_HAVE_QUICKJS_ADD_INTRINSIC_BIG_INT)                           \
+    && !NJS_HAVE_QUICKJS_BASE_OBJECTS_BIG_INT
```
bnoordhuis: This won't be necessary after https://github.com/quickjs-ng/quickjs/pull/1670 lands.

xeioex: @bnoordhuis 

Thanks. Do you plan to make a new tag for this?

VadimZhestikov (approved):
Looks good

  Minor observations (non-blocking)

  1. The 128 offset is heuristic, not future-proof. It's JS_CLASS_INIT_COUNT (67)
     plus headroom, and the comment says as much. If a future quickjs-ng grows
     JS_CLASS_INIT_COUNT past 128, or heavy dynamic class allocation reaches 128
     before njs registers, the collision returns. It's the one thing to re-check on
     future engine bumps. A compile-time guard would be ideal but JS_CLASS_INIT_COUNT
     isn't a stable public constant, so the pragmatic fixed offset is acceptable.
  2. Commit 2 also changes behavior on older engines (it removed the version gate),
     now always using the JS_NewObjectClass path. It's sound by construction (the
     JS_NewError == JS_NewObjectClass + stack invariant holds across versions) and CI
     covers older engines — I verified it on 0.16.2 but didn't run an older build.
  3. External-buffer + transfer() at the same length moves the pool-backed pointer
     into a new ArrayBuffer (engine won't free it — correct). The "retained past its
     pool" lifetime concern there is the pre-existing external-buffer class (the
     #1112/N2 territory), not introduced by this PR.

PR #1116, opened 2026-08-22
String: fix out-of-bounds read in $' substitution expansion

basavaraj-sm05:
### Proposed changes

The `$'` token in a replacement string expands to the part of the subject that comes after the match. `njs_string_get_substitution()` builds it with `njs_string_offset(&s, pos + length)`, where `length` is the character length of the matched text. `pos` is already clamped to the subject by every caller, but the matched length is taken as-is, and through `RegExp.prototype[Symbol.replace]` with a script-supplied `exec()` the match object's `[0]` can be any string, so `pos + length` can run well past the end of the subject. On a multibyte subject that pushes `njs_string_offset()` into `njs_string_utf8_offset()`, whose own comment notes it assumes a valid index and which reads the UTF-8 offset map without a bounds check, so the oversized index reads far outside the map. I noticed it while comparing the `$'` branch against the neighboring `` $` `` branch, which stays in range because it only uses the already-clamped `pos`. Clamping `pos + length` to the subject length before the offset call closes the read and matches the GetSubstitution step in the spec, where the tail starts at `min(position + matchLength, stringLength)`.

### Checklist

### DIFF
--- a/src/njs_string.c
+++ b/src/njs_string.c
@@ -3000,7 +3000,9 @@ njs_string_get_substitution(njs_vm_t *vm, njs_value_t *matched,
             length = njs_string_prop(vm, &m, matched);
             (void) njs_string_prop(vm, &s, string);
 
-            tail = njs_string_offset(&s, pos + length) - s.start;
+            tail = njs_string_offset(&s, njs_min(pos + (int64_t) length,
+                                                 (int64_t) s.length))
+                   - s.start;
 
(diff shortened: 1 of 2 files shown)


xeioex:
Hi @basavaraj-sm05,

Thank you for the report. The issue is valid, but I prefer a proper fix which prevents the problem found in this PR by construction.

PR #1117, opened 2026-08-28
fix off-by-one write in c14n namespace prefix list

basavaraj-sm05:
### Proposed changes

The C14N helpers (`xml.c14n`, `xml.exclusiveC14n`, `xml.serializeToString`) take an inclusive-namespace `prefixes` string straight from script and split it into a NULL-terminated pointer array in `njs_xml_parse_ns_list()`. The array starts at eight slots and doubles whenever the next token would not fit, but the growth check only reserves room for the token itself, never for the terminating NULL. So once the token count reaches the current capacity exactly (8, 16, 32, ...), every slot is filled and the final `*out = NULL` writes one element past the end of the allocation. I ran into it while exercising `exclusiveC14n` with longer prefix lists: at 128 tokens the array is a 1 KiB pool allocation and ASAN reports an 8-byte heap-buffer-overflow write at the terminator. Growing one slot earlier (`idx + 1 >= size`) keeps a spare slot for the terminator and leaves valid-input behavior unchanged. The QuickJS module carries the same helper with the same off-by-one, so I fixed both, and the added c14n case uses a power-of-two prefix count to cover the boundary the existing ten-token test steps over.

### Checklist


xeioex:
@basavaraj-sm05 

Thank you,

This was reported earlier and fixed as a security fix in 1.0.1 njs release.

PR #1120, opened 2026-09-04
Fix the runtime model in the agent developer guide

xeioex:


VadimZhestikov (approved):
An accurate, well-targeted correction of a guide that previously stated several outright-wrong things about the  runtime model.

PR #1123, opened 2026-09-09
TypedArray: coerce length argument once on buffer construction

basavaraj-sm05:
### Proposed changes

Constructing a typed array from an ArrayBuffer with an explicit length, such as `new Int8Array(buffer, byteOffset, length)`, runs the length through `njs_value_to_index()` twice in `njs_typed_array_alloc()`: once before the detached-buffer check and once after it, keeping only the second result. Because `ToIndex` evaluates the argument's `valueOf()`, the length object's `valueOf()` fires twice, which is observable from script and departs both from the specification, where `ToIndex` runs once, and from the `DataView` constructor, which converts its length a single time. The duplicate conversion also runs after the detached-buffer check, so a `valueOf()` that detaches the buffer while the length is being coerced slips past the check and the constructor returns a zero-length view over the detached buffer instead of throwing. The fix removes the second conversion so the length is coerced once, before the check. A `valueOf()` that detaches mid-coercion is now reported, and valid construction is unchanged. I added a regression test asserting the length argument's `valueOf()` is invoked exactly once.

### Checklist


xeioex (approved):
Thank you for the patch.

I simplified the commit log because it had factual inaccuracies.

Otherwise LGTM.

Please do The F5 CLA, so I will be able to merge.

PR #1124, opened 2026-09-10
Xml modify in place

xeioex:
Previously, text and structural mutators copied and replaced the receiver's libxml2 node.  Existing XMLNode wrappers continued to point at the detached original, while subsequent lookups returned the replacement.  Wrappers for
children preserved by the operation likewise pointed into the detached old subtree.  In QuickJS, mutating the detached receiver again could dereference the NULL result of xmlReplaceNode().

The fix is to modify the existing receiver's child list directly.  Existing wrappers and subsequent lookups therefore continue to refer to the same libxml2 node.  Removed child lists remain readable until the QuickJS document or njs VM is destroyed.  A node is considered live only when following its parent links reaches its xmlDoc.  Replacement children and cleanup records are prepared before the live tree is modified.

Changed behavior:
- A removed child has an undefined $parent, while its descendants retain
  their internal parent links.
- Mutating or canonicalizing a removed XMLNode throws TypeError.
- Text and attribute values must be primitive strings; null and undefined
  retain their deletion semantics.
- C14N rejects an excluding node from another document or a detached list.
- addChild() and $tags reject namespace-bearing or unsupported source trees
  before mutation.
- Named $tags assignments are consistently rejected by both engines.

While here, use an unsigned QuickJS $tags array length, support assigning $tags on an empty njs XMLNode, and make deleting $tags from an empty node a no-op.


VadimZhestikov (dismissed):
Looks good.

  Notes (non-blocking)

  1. Memory tradeoff: removed subtrees are retained until the document/VM is destroyed (the retired list). This is a
     deliberate correctness-over-memory choice and is fine for nginx's request-scoped documents, but a script that
     repeatedly removes children from a long-lived document accumulates memory until teardown. Worth a one-line mention
     in docs; not a leak (it's freed at doc destruction).
  2. The two engines carry parallel implementations of the same redesign (+311/−233 njs, +319/−174 QuickJS); the test
     suite exercises both, and the behavior-parity items (e.g. named $tags rejection) are explicitly tested on each.
  3. The NULL-xmlReplaceNode crash is eliminated structurally (the function is gone), so I didn't need to reproduce the
     exact SEGV — the base's silent mishandling of the detached receiver in the negative control already shows the old
     model was unsafe.

VadimZhestikov (approved):
Looks good

PR #1125, opened 2026-09-12
NJS test262 runner

xeioex:
Adds `build/njs_test262`, a native runner for the official TC39 test262 suite.  Unlike the existing `make test262` shell harness, it embeds njs directly and executes the upstream suite without spawning one CLI process per test.

The runner:

- parses test262 frontmatter and assembles harness sources;
- supports strict scripts, modules, async tests, negative tests, feature filtering, exclusions, and targeted paths;
- reports PASS, FAIL, SKIP, CRASH, and INFRA results;
- supports process isolation and a per-variant CPU limit;
- compares failures against a sorted path/variant baseline, reporting NEW, FIXED, and STALE entries;
- refuses baseline updates after crashes or infrastructure errors.

The official suite is pinned in `test/njs_test262.version`; the configured profile and expected failures are stored in `test/njs_test262.conf` and `test/njs_test262.failures`.

A dedicated `test262` GitHub Actions workflow runs the full profile in normal, ASan, and UBSan builds.

```
variants: 31951, pass: 14669, fail: 1869, skip: 15413, crash: 0, infra: 0, duration: 32.250s
new: 0, fixed: 0
```

### DIFF
--- a/src/njs_atom.c
+++ b/src/njs_atom.c
@@ -235,9 +235,8 @@ njs_atom_atomize_key(njs_vm_t *vm, njs_value_t *value)
     switch (value->type) {
     case NJS_STRING:
         num = njs_key_to_index(value);
-        u32 = (uint32_t) num;
 
-        if (njs_fast_path(u32 == num && (u32 < 0x80000000)
+        if (njs_fast_path(njs_number_is_atom_index(num, &u32)
                           && !(num == 0 && signbit(num))))
         {
             value->atom_id = njs_number_atom(u32);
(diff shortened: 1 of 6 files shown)


VadimZhestikov (dismissed):
Looks good

  Notes (non-blocking)

  1. NJS_CHB_CTX_INIT still casts js_malloc/js_free to the CHB callback types (same incompatible-function-pointer UB pattern that "CHB incompatible function pointers (cf9c468c)" fixes for the mp variant). It's QuickJS-path and wasn't flagged (GCC doesn't instrument -fsanitize=function; the njs UBSan CI wouldn't catch it), but it's the same latent UB — worth a follow-up adapter for consistency.
  2. The pinned test262 version (419d3e0a) and njs_test262.failures are version-locked; running against a different checkout produces expected drift in the failure set. Not an issue — the CI (test262.yml) would use the pinned version.
  3. These are UB/correctness fixes rather than exploitable memory-safety bugs (except the alignment one, which is genuine UB on strict-alignment targets); the Date UTC-setter fix is user-visible and worth a CHANGES note.

VadimZhestikov (approved):
Looks good

PR #1126, opened 2026-09-16
Fetch hardening

xeioex:
This patch set hardens Fetch response handling and js_periodic parsing:
- Rejects negative max_response_body_size values instead of returning an apparently successful empty response.
- Validates response framing by rejecting conflicting Content-Length fields and responses containing both Transfer-Encoding and Content-Length.
- Handles response framing explicitly: HEAD, 204, and 304 responses have no body; fixed-length responses finish at Content-Length; close-delimited responses finish at EOF; chunked responses finish at the terminal chunk.
- Rejects unsupported informational 1xx responses instead of returning an incorrect interim response.
- Prevents reuse of keepalive connections with already-buffered trailing response data.
- Applies max_response_body_size to decoded chunked bodies, replacing the previous tenfold limit.

- Fixes worker_affinity parsing when it precedes interval or jitter.


VadimZhestikov (approved):
Looks good

Notes (non-blocking)

  1. Identical duplicate Content-Length is accepted (only differing values are rejected). That's RFC 7230 §3.3.2-compliant, but some hardened clients reject any duplicate CL. Defensible as-is; worth a conscious decision since duplicate-CL tolerance is occasionally itself a smuggling nuance.
  2. Threat framing: these are all response-side protections, so the adversary is a malicious or compromised upstream that an njs script fetches from. Relevant whenever ngx.fetch targets untrusted/semi-trusted upstreams; worth a CHANGES/security note, and the keepalive-poisoning one "Avoid reusing framed response connections (6ac531ca)" is the most impactful (cross-request confusion in the  shared fetch keepalive cache).

PR #1128, opened 2026-09-18
Memory pool fixes

xeioex:
This series adds direct test coverage for the memory pool and applies several independent allocator improvements.
- Adds mp_unit_test covering allocation, reuse, alignment, mixed allocation sizes, multiple valid pool configurations, runtime page-size configuration, churn, and stress scenarios.
- Runs library tests from the standard make test target.
- Adds an ASan CI configuration without debug memory, so normal page and cluster allocator paths are covered.
- Fixes library-test dependency generation for the new executable.
- Avoids undefined signed shifts in MurmurHash2 under UBSan.
- Fixes 32-bit rbtree address comparison overflow that could make a pool remain non-empty after all allocations were freed.
- Restricts freed-memory junk filling to debug builds.
- Replaces per-page cluster scans with an O(1) free-page counter.
- Caches the chunk slot index in each page descriptor to avoid a free-side slot search.


VadimZhestikov (approved):
Looks good

  Notes (non-blocking)

  - N4 removes the incidental 0x5A poisoning of freed memory from release builds — a minor loss of a UAF-debugging aid
    (kept under NJS_DEBUG; ASan/valgrind remain the real detectors). Worth a one-line mention, not a concern.
  - N2 also wires the library tests (flathsh/rbtree/mp) into the standard test target + a UBSan CI job — which is
    exactly what surfaced N1; good hygiene, and it means these paths are now sanitizer-covered in CI.

